PR #3, opened 2016-02-08
Adding note about type error for sets (and dicts), and allow custom --gen

tbug:
Had issue compiling the the generated erlang due to these two errors

```
type dict() undefined
type set() undefined
```
The dict and set types where non-namespaced in the generated hrl (`set()` instead of `sets:set()`) so I'm guessing the default thrift erl options is `otp16`.

From the `thrift -help`:
```
  erl (Erlang):
    legacynames: Output files retain naming conventions of Thrift 0.9.1 and earlier.
    maps:        Generate maps instead of dicts.
    otp16:       Generate non-namespaced dict and set instead of dict:dict and sets:set.
```

Most people probably want `maps` instead.
I added a note about the type error and suggests a custom `thrift_options`,
and made sure that `--gen erl` is only added to the system call when `--gen` is not in the `thrift_options` to allow people to set it themselves.

### DIFF
--- a/lib/mix/tasks/compile.thrift.ex
+++ b/lib/mix/tasks/compile.thrift.ex
@@ -46,10 +46,10 @@ defmodule Mix.Tasks.Compile.Thrift do
         Mix.raise "Unsupported Thrift version #{v} (requires #{thrift_version})"
       end
     end
-
     unless(Enum.empty?(stale_files), do: File.mkdir_p!(output_dir))
 
-    Enum.each stale_files, &generate(&1, output_dir, thrift_options)
+    options = get_thrift_options(output_dir, thrift_options)
+    Enum.each stale_files, &generate(&1, options)
   end
 
   defp get_thrift_version do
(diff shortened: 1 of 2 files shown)


lib/mix/tasks/compile.thrift.ex
```diff
@@ -71,7 +71,12 @@ defmodule Mix.Tasks.Compile.Thrift do
   end
 
   defp generate(thrift_file, output_dir, options) do
-    args = ~w[--gen erl --out] ++ [output_dir] ++ options ++ [thrift_file]
+    # if --gen is in the options given, leave it out from the base args
+    baseargs = case Enum.member?(options, "--gen") do
+      true  -> ~w[--out] ++ [output_dir]
+      false -> ~w[--gen erl --out] ++ [output_dir]
+    end
+    args = baseargs ++ options ++ [thrift_file]
```
jparise: `generate` is called once for each Thrift file. It would be better to do this work earlier (in `run`) and pass a full set of options to generate. A `get_thrift_options` private helper routine would make the code pretty clear.

`generate` would then just append the filename and invoke the `thrift` binary.

tbug: Indeed, although, performance is probably a non-issue in this mix task :smile:
I'll clean it up :+1:

lib/mix/tasks/compile.thrift.ex
```diff
@@ -70,8 +70,16 @@ defmodule Mix.Tasks.Compile.Thrift do
     Enum.empty?(targets) || Mix.Utils.stale?([thrift_file], targets)
   end
 
-  defp generate(thrift_file, output_dir, options) do
-    args = ~w[--gen erl --out] ++ [output_dir] ++ options ++ [thrift_file]
+  defp get_thrift_options(output_dir, user_options) do
+    opts = ~w[--out] ++ [output_dir]
+    # add --gen erl if user options doesn't contain --gen
+    opts = opts ++ if Enum.member?(user_options, "--gen"), do: ~w[--gen erl]
```
jparise: Is this logic correct? It appears to add `--gen erl` if `--gen` appears in the user-provided options, which seems to be the opposite of what we want.

tbug: Ah, no, it isn't.
My bad.
I should stop coding before the first cup of coffee.

jparise: Just changing the `if` to `unless` should do it.

jparise:
@tbug: This looks great. Thanks for the improvements!

Before I merge, would you mind squashing your three commits into one?

tbug:
Yes, ofc. Sorry for the mess :smile:

jparise:
Thanks!

PR #6, opened 2016-06-09
An implementation of the Thrift parser

scohen:
This PR introduces a parser based on @jparise's lexer. The parser
runs on thrift files and produces a set of Program models. The
intention is to be able to give Elixir full-fledged thrift support
and a better (and faster) implementation.

@jparise @stemmler

### DIFF
--- a/lib/parser/models.ex
+++ b/lib/parser/models.ex
@@ -1,0 +1,473 @@
+  defmodule Thrift.Parser.Types do
+    defmodule Primitive do
+      @type t :: :bool | :i8 | :i16 | :i64 | :binary | :double | :byte | :string
+    end
+
+    defmodule Ident do
+      @type t :: String.t
+    end
+
+    defmodule Standalone do
+      @type t :: Ident.t | Primitive.t
+    end
+
+    defmodule List do
+      @type t :: {:list, Thrift.Parser.Types.t}
+    end
+
+    defmodule Map do
+      @type t  :: {:map, {Thrift.Parser.Types.t, Thrift.Parser.Types.t}}
+    end
+
+    defmodule Set do
+      @type t :: {:set, Thrift.Parser.Types.t}
+    end
+
+    defmodule Container do
+      @type t :: List.t | Map.t | Set.t
+    end
+
+    @type t :: Container.t | Standalone.t
+  end
+
+  defmodule Thrift.Parser.Literals do
+    defmodule Primitive do
+      @type t :: integer | boolean | String.t | float
+    end
+
+    defmodule List do
+      @type t :: [Thrift.Parser.Literals.t]
+    end
+
+    defmodule Map do
+      @type t :: %{Thrift.Parser.Literals.t => Thrift.Parser.Literals.t}
+    end
+
+    defmodule Container do
+      @type t :: Map.t | List.t
+    end
+
+    @type t :: Container.t | Primitive.t
+    @type s :: atom
+  end
+
+  defmodule Thrift.Parser.Conversions do
+    def atomify(nil), do: nil
+    def atomify(l) when is_list(l) do
+      List.to_atom(l)
+    end
+
+    def cast(_, nil) do
+      nil
+    end
+
+    def cast(:double, val) do
+      val
+    end
+
+    def cast(:string, val) do
+      List.to_string(val)
+    end
+
+    def cast({:set, type}, val) do
+      MapSet.new(val, &cast(type, &1))
+    end
+
+    def cast({:map, {key_type, val_type}}, val) do
+      Enum.into(val, %{}, fn {k, v} ->
+        {cast(key_type, k), cast(val_type, v)}
+      end)
+    end
+
+    def cast(_, val) do
+      val
+    end
+  end
+
+  defmodule Thrift.Parser.Models do
+    alias Thrift.Parser.Types
+    alias Thrift.Parser.Literals
+
+    defmodule Namespace do
+      @moduledoc """
+      A Thrift namespace.
+      The namespace is a language-specific place where the generated structs are
+      placed.
+      """
+
+      @type t :: %Namespace{name: String.t, path: String.t}
+      defstruct name: nil, path: nil
+
+      import Thrift.Parser.Conversions
+
+      @spec new(char_list, char_list) :: %Namespace{}
+      def new(name, path) do
+        %Namespace{name: atomify(name), path: List.to_string(path)}
+      end
+    end
+
+    defmodule Include do
+      @moduledoc """
+      An included file.
+      In Thrift, you can include other files to share structs, enums and the like.
+      """
+
+      @type t :: %Include{path: String.t}
+      defstruct path: nil
+
+      import Thrift.Parser.Conversions
+
+      @spec new(char_list) :: %Include{}
+      def new(path) do
+        %Include{path: List.to_string(path)}
+      end
+    end
+
+    defmodule Constant do
+      @moduledoc """
+      A Thrift constant.
+      Constants of any primitive or container type can be created in Thrift.
+      """
+
+      @type t :: %Constant{name: String.t, value: Literal.t, type: Types.t}
+      defstruct name: nil, value: nil, type: nil
+
+      import Thrift.Parser.Conversions
+
+      @spec new(char_list, Literals.t, Types.t) :: %Constant{}
+      def new(name, val, type) do
+        %Constant{name: atomify(name), value: cast(type, val), type: type}
+      end
+    end
+
+    defmodule TEnum do
+      @moduledoc """
+      A Thrift enumeration
+      An enumeration contains names and (usually sequential) values, and
+      allows you to map from one to the other.
+      """
+
+      @type enum_value :: bitstring | integer
+      @type t :: %TEnum{name: String.t, values: %{String.t => enum_value}}
+      defstruct name: nil, values: []
+
+      import Thrift.Parser.Conversions
+
+      @spec new(char_list, %{char_list => enum_value}) :: %TEnum{}
+      def new(name, values) do
+        values = values
+        |> Enum.with_index
+        |> Enum.map(fn
+          {{name, value}, _index} ->
+            {atomify(name),  value}
+
+          {name, index} ->
+            {atomify(name), index + 1}
+        end)
+
+        %TEnum{name: atomify(name), values: values}
+      end
+    end
+
+    defmodule Field do
+      @moduledoc """
+      A Thrift field.
+
+      Fields define a named type and can occur in functions, structs, unions,
+      exceptions and the parameter list and `throws` clauses of functions.
+
+      Fields can refer to each other. These are represented by the FieldReference
+      type.
+
+      This module also contains some utilities for validating and fixing up fields.
+      """
+
+      @type printable :: String.t | atom
+      @type t :: %Field{id: integer, name: String.t, type: Types.t,
+                        required: boolean, default: Literals.t}
+      defstruct id: nil, name: nil, type: nil, required: :default, default: nil
+
+      import Thrift.Parser.Conversions
+
+      @spec new(integer, boolean, Types.t, char_list, Literals.t) :: %Field{}
+      def new(id, required, type, name, default) do
+        %Field{id: id,
+               type: type,
+               name: atomify(name),
+               required: required,
+               default: cast(type, default)}
+      end
+
+      @spec build_field_list(printable, [%Field{}]) :: [%Field{}]
+      def build_field_list(parent_name, fields) do
+        fields
+        |> update_ids(parent_name)
+        |> validate_ids(parent_name)
+      end
+
+      defp validate_ids(fields, name) do
+        dupes = fields
+        |> Enum.group_by(&(&1.id))
+        |> Enum.filter(fn {_, v} -> length(v) > 1 end)
+
+        unless Enum.empty?(dupes) do
+          {id, fields} = List.first(dupes)
+
+          names = fields
+          |> Enum.map(&("#{name}.#{&1.name}"))
+          |> Enum.join(", ")
+
+          raise "Error: #{names} share field number #{id}."
+        end
+
+        fields
+      end
+
+      defp update_ids(fields, parent_name) do
+        alias Thrift.Parser.Shell
+        fields
+        |> Enum.with_index
+        |> Enum.map(fn
+          {field=%__MODULE__{}, idx} ->
+            case field.id do
+              nil ->
+                Shell.warn "Warning: id not set for field '#{parent_name}.#{field.name}'."
+                %__MODULE__{field | id: idx + 1}
+              _ ->
+                field
+            end
+        end)
+      end
+    end
+
+    defmodule Exception do
+      @moduledoc """
+      A Thrift exception
+
+      Exceptions can happen when the remote service encounters an error.
+      """
+
+      @type t :: %Exception{name: String.t, fields: [%Field{}]}
+      defstruct fields: %{}, name: nil
+
+      import Thrift.Parser.Conversions
+      alias Thrift.Parser.Models.Field
+
+      @spec new(char_list, [%Field{}, ...]) :: %Exception{}
+      def new(name, fields) do
+        ex_name = atomify(name)
+        updated_fields = Field.build_field_list(ex_name, fields)
+
+        %Exception{name: ex_name, fields: updated_fields}
+      end
+    end
+
+    defmodule Struct do
+      @moduledoc """
+      A Thrift struct
+
+      The basic datastructure in Thrift, structs have aa name and a field list.
+      """
+
+      @type t :: %Struct{name: String.t, fields: %{String.t => %Field{}}}
+      defstruct name: nil, fields: %{}
+
+      import Thrift.Parser.Conversions
+      alias Thrift.Parser.Models.Field
+
+      @spec new(char_list, [%Field{}, ...]) :: %Struct{}
+      def new(name, fields) do
+        struct_name = atomify(name)
+        fields = Field.build_field_list(struct_name, fields)
+
+        %Struct{name: struct_name, fields: fields}
+      end
+    end
+
+    defmodule Union do
+      @moduledoc """
+      A Thrift union
+
+      Unions can have one field set at a time.
+      """
+
+      @type t :: %Union{name: String.t, fields: %{String.t => %Field{}}}
+      defstruct name: nil, fields: %{}
+
+      import Thrift.Parser.Conversions
+      alias Thrift.Parser.Models.Field
+
+      @spec new(char_list, [%Field{}, ...]) :: %Union{}
+      def new(name, fields) do
+        name = atomify(name)
+        fields = Field.build_field_list(name, fields)
+        |> Enum.map(fn(field=%Field{}) ->
+          # According to Thrift docs, unions have implicitly optional
+          # fields. See https://thrift.apache.org/docs/idl#union
+          %Field{field | required: false}
+        end)
+
+        %Union{name: name, fields: fields}
+      end
+    end
+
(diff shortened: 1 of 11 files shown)


jparise:
`src/thrift_parser.erl` should be `.gitignore`'ed. We want to locally (re)generate that output file as part of the build.

jparise:
src/thrift_parser.erl should be .gitignore'ed. We want to locally (re)generate that output file as part of the build.

lib/parser/models.ex
```diff
@@ -0,0 +1,385 @@
...
+      @moduledoc """
+      A thrift namespace.
+      The namespace is a language-specific place where the generated structs are
+      placed.
+      """
+      defstruct name: nil, path: nil
+      import Thrift.Parser.Conversions
+
+      def new(name, path) do
+        %Namespace{name: atomify(name), path: stringify(path)}
```
jparise: Does this function (and the similar ones below) take variably-typed input values? I'm trying to better understand the need for generic `atomify` and `stringify` helpers.

lib/parser/models.ex
```diff
@@ -0,0 +1,385 @@
...
+      def new(name, values) do
+        values = values
+        |> Enum.with_index
+        |> Enum.map(fn
+          {{name, value}, _index} ->
+            {atomify(name),  value}
+
+          {name, index} ->
+            {atomify(name), index + 1}
+
```
jparise: Unnecessary newline.

lib/parser/models.ex
```diff
@@ -0,0 +1,385 @@
...
+      def new(name, val, type) do
+        %Constant{name: atomify(name), value: cast(type, val), type: type}
+      end
+
+    end
+
+    defmodule TEnum do
+      @moduledoc """
+      A Thrift enumeration
+      An enumeration contains names and (usually sequential)values and
```
jparise: Missing space after `)`.

lib/parser/models.ex
```diff
@@ -0,0 +1,385 @@
...
+
+        end)
+
+        %TEnum{name: atomify(name), values: values}
+      end
+    end
+
+    defmodule Field do
+      @moduledoc """
+      A thrift field.
```
jparise: Capitalize "thrift" consistently throughout the documentation.

lib/parser/models.ex
```diff
@@ -0,0 +1,385 @@
...
+
+      Fields define a named type and can occur in functions, structs, unions,
+      exceptions and the parameter list and throws clauses of functions.
```
jparise: Markup `throws` as a literal?

lib/parser/models.ex
```diff
@@ -0,0 +1,385 @@
...
+      This module also contains some utilities for validating and fixing up fields.
+      """
+      defstruct id: nil, name: nil, type: nil, required: :default, default: nil
+      import Thrift.Parser.Conversions
+
+      def new(id, required, type, name, default) do
+
+        casted_default = case default do
+                           nil ->
+                             nil
```
jparise: Would this be simpler if we defined `cast(_, nil) -> nil`?

lib/parser/models.ex
```diff
@@ -0,0 +1,385 @@
...
+
+      defp update_ids(fields, parent_name) do
+        alias Mix.Shell.IO
+        fields
+        |> Enum.with_index
+        |> Enum.map(fn
+          {field=%__MODULE__{}, idx} ->
+            case field.id do
+              nil ->
+                IO.info "Warning: id not set for field '#{parent_name}.#{field.name}'."
```
jparise: Should this `raise` or do something that doesn't produce explicit IO?

lib/parser/models.ex
```diff
@@ -0,0 +1,385 @@
...
+          |> Enum.join(", ")
+
+          raise "Error: #{names} share field number #{id}."
+        end
+
+        fields
+      end
+
+      defp update_ids(fields, parent_name) do
+        alias Mix.Shell.IO
```
jparise: No guarantee that we'll be run from Mix, right?

lib/parser/models.ex
```diff
@@ -0,0 +1,385 @@
...
+                %__MODULE__{field | id: idx + 1}
+              _ ->
+                field
+            end
+        end)
+      end
+
+    end
+
+
```
jparise: Unnecessary newline.

lib/parser/models.ex
```diff
@@ -0,0 +1,385 @@
...
+    defmodule Exception do
+      @moduledoc """
+      A thrift exception
+
+      """
```
jparise: More documentation needed here?

lib/parser/models.ex
```diff
@@ -0,0 +1,385 @@
...
+      import Thrift.Parser.Conversions
+      alias Thrift.Parser.Models.Field
+
+      def new(name, fields) do
+        name = atomify(name)
+        fields = Field.build_field_list(name, fields)
+        |> Enum.map(fn(field=%Field{}) ->
+          # According to thrift docs, unions have implicitly optional
+          # fields. See https://thrift.apache.org/docs/idl#union
+          %Field{field | required: false}
```
jparise: Could this also use `put_in`?

```elixir
|> put_in([:required], false)
```

Untested and not sure if that works the same way with structs, but just curious.

lib/parser/models.ex
```diff
@@ -0,0 +1,385 @@
...
+    end
+
+    defmodule StructRef do
+      @moduledoc """
+      A reference to another struct.
+
+      While not a thrift type, this represents when a thrift type refers to
+      another.
+      """
+
```
jparise: Unnecessary newline.

lib/parser/models.ex
```diff
@@ -0,0 +1,385 @@
...
+        name = atomify(name)
+        fields = Field.build_field_list(name, fields)
+        |> Enum.map(fn(field=%Field{}) ->
+          # According to thrift docs, unions have implicitly optional
+          # fields. See https://thrift.apache.org/docs/idl#union
+          %Field{field | required: false}
+        end)
+        %Union{name: name, fields: fields}
+      end
+
```
jparise: Unnecessary newline.

lib/parser/models.ex
```diff
@@ -0,0 +1,385 @@
...
+      A thrift service
+
+      Services hold RPC functions and can extend other services.
+      """
+      defstruct name: nil, extends: nil, functions: []
+
+      import Thrift.Parser.Conversions
+
+      def new(name, functions, extends) do
+
```
jparise: Unnecessary newline.

lib/parser/models.ex
```diff
@@ -0,0 +1,385 @@
...
+      import Thrift.Parser.Conversions
+      alias Thrift.Parser.Models.Include
+      alias Thrift.Parser.Models.Namespace
+      alias Thrift.Parser.Models.Constant
+      alias Thrift.Parser.Models.Exception
+      alias Thrift.Parser.Models.Struct
+      alias Thrift.Parser.Models.TEnum
+      alias Thrift.Parser.Models.Union
+
+      def new(nil, defs) do
```
jparise: Define `defs`?

lib/parser/models.ex
```diff
@@ -0,0 +1,385 @@
...
+        new([], defs)
+      end
+
+      def new(headers, nil) do
```
jparise: Define `headers`?

mix.exs
```diff
@@ -58,6 +58,8 @@ defmodule Thrift.Mixfile do
       files: ~w(README.md LICENSE mix.exs lib) ++
              ~w(ext/thrift/CHANGES ext/thrift/LICENSE ext/thrift/NOTICE) ++
              ~w(ext/thrift/README.md ext/thrift/doc ext/thrift/lib/erl) ++
-             ~w(src/thrift_lexer.xrl)]
+             ~w(src/thrift_lexer.xrl) ++
+             ~w(src/thrift_parser.yrl)
```
jparise: I think this would look nicer on one line:

```elixir
~w(src/thrift_lexer.xrl src/thrift_parser.yrl)
```

src/thrift_lexer.xrl
```diff
@@ -7,23 +7,46 @@ HEX             = [+-]?0x[0-9A-Fa-f]+
 DOUBLE          = [+-]?[0-9]+(\.[0-9]+)?([eE][+-]?[0-9]+)?
 IDENTIFIER      = [a-zA-Z_](\.[a-zA-Z_0-9]|[a-zA-Z_0-9])*
 ST_IDENTIFIER   = [a-zA-Z-](\.[a-zA-Z_0-9-]|[a-zA-Z_0-9-])*
-SYMBOL          = [:;\,\{\}\(\)\=<>\[\]]
+SEMI            = [;]
```
jparise: `SEMICOLON` might be a little clearer.

lib/parser/models.ex
```diff
@@ -0,0 +1,385 @@
...
+          oneway: oneway,
+          return_type: return_type,
+          name: name,
+          params: params,
+          exceptions: exceptions
+        }
+      end
+    end
+
+    defmodule Program do
```
jparise: I wonder if this would be better named "Schema" or "Grammar". What was the thinking behind "Program"?

src/thrift_parser.yrl
```diff
@@ -0,0 +1,238 @@
+
+Nonterminals
```
jparise: I sort of like how the Elixir grammar groups these into lines of multiple names:

```
Nonterminals
  program
  headers header
  definitions definition
...
```

src/thrift_parser.yrl
```diff
@@ -0,0 +1,238 @@
...
+
+field_default -> '$empty': nil.
+field_default -> equals literal: '$2'.
+
+typedef_def ->
+    typedef type ident:
+        {typedef, '$2', unwrap('$3')}.
+
+struct_def ->
+    struct ident open_curly fields close_curly:
```
jparise: This might read better with the literal `'{`'` and `'}'` characters. What are the tradeoffs between that and using tokens from the lexer?

test/parser_test.exs
```diff
@@ -0,0 +1,501 @@
...
+      i32 count,
+      i64 long_thing = 12345
+      optional list<i32> optional_list,
+    }
+    """
+
+    warnings = capture_io(fn ->
+      s = parse(struct_def, [:structs, :Optionals])
+
+      assert s == %Struct{
```
jparise: I would drop the `s` and just do:

```elixir
assert parse(...) == %Struct{
```

lib/parser/models.ex
scohen: doh, that would work.

lib/parser/models.ex
scohen: This is what the current thrift implementation does; it's only a warning.

lib/parser/models.ex
scohen: but I wanted colors!

lib/parser/models.ex
scohen: I have to talk about exceptions? Ok...

lib/parser/models.ex
scohen: It can't, I get an error about this not supporting the Access protocol. Map.put works, but I think this is the preferred method for updating structs.

jparise: That makes sense. I wasn't able to try it myself, so I'm glad we're already using the best pattern here.

src/thrift_parser.yrl
scohen: I just used the lexer's defined symbols. I'll give it a shot with the literals.

test/parser_test.exs
scohen: I kinda like the line breaks. When developing, it helped me put intermediate stuff in there.

src/thrift_parser.yrl
scohen: I'll have to change how the lexer... lexes. I think it's worth it though.

lib/parser/models.ex
scohen: I'm pretty sure we atomify things that are already atoms (like nil), but I can safely remove the bitstring versions. 

Don't the guard clauses tell the story here? What would a spec add?

jparise: Sorry, my comment was ambiguous. I mean specs for the `new` functions within these struct modules.

lib/parser/models.ex
```diff
@@ -0,0 +1,479 @@
...
+      A Thrift namespace.
+      The namespace is a language-specific place where the generated structs are
+      placed.
+      """
+      defstruct name: nil, path: nil
+      import Thrift.Parser.Conversions
+
+      @spec new(char_list, char_list) :: %Namespace{}
+      def new(name, path) do
+        %Namespace{name: atomify(name), path: stringify(path)}
```
jparise: I think this now reenforces my earlier thought that it would be nicer to just use `List.to_atom` and `List.to_string` directly. I find those explicit conversions easier to reason about.

That would let us drop `stringify` entirely because it's only used here an in `Include.new/1`.

lib/parser/models.ex
```diff
@@ -0,0 +1,451 @@
...
+      @spec new(integer, boolean, TTypes.thrift_type, char_list, TTypes.literal) :: %Field{}
+      def new(id, required, type, name, default) do
+
+        casted_default = cast(type, default)
+
+        %Field{id: id,
+               type: type,
+               name: atomify(name),
+               required: required,
+               default: casted_default}
```
jparise: Just write `cast(type, default)` here as you do with the `atomify` call?

lib/parser/models.ex
```diff
@@ -0,0 +1,451 @@
...
+        |> Enum.filter(fn {_, v} -> length(v) > 1 end)
+
+        unless Enum.empty?(dupes) do
+          {id, fields} = List.first(dupes)
+
+          names = fields
+          |> Enum.map(&("#{name}.#{&1.name}"))
+          |> Enum.join(", ")
+
+          raise "Error: #{names} share field number #{id}."
```
jparise: Why not `Shell.error` here?

lib/parser/models.ex
```diff
@@ -0,0 +1,451 @@
...
+    end
+
+    defmodule StructRef do
+      @moduledoc """
+      A reference to another struct.
+
+      While not a Thrift type, this represents when a Thrift type refers to
+      another.
+      """
+
```
jparise: Extra newine.

src/thrift_parser.yrl
```diff
@@ -0,0 +1,206 @@
...
+
+constant_def ->
+    const type ident '=' literal:
+        'Elixir.Thrift.Parser.Models.Constant':new(unwrap('$3'), '$5', '$2').
+
+ns_name -> '*': "*".
+ns_name -> ident: unwrap('$1').
+
+
+%% JS Style mapping "foo": 32
```
jparise: "JS"? Is this a JavaScript reference?

test/parser_test.exs
```diff
@@ -0,0 +1,554 @@
...
+  alias Thrift.Parser.Models.Namespace
+  alias Thrift.Parser.Models.Service
+  alias Thrift.Parser.Models.Struct
+  alias Thrift.Parser.Models.StructRef
+  alias Thrift.Parser.Models.TEnum
+  alias Thrift.Parser.Models.Union
+
+  import ExUnit.CaptureIO
+
+  defp parse(doc, path) do
```
jparise: Should this function (and the one below) be part of the library's public API (in `lib/parser.ex`)?

test/parser_test.exs
```diff
@@ -0,0 +1,554 @@
...
+    {:ok, parse_tree} = :thrift_parser.parse(tokens)
+
+    parse_tree
+  end
+
+  test "parsing comments" do
+    """
+    // a simple C-style comment
+    """
+    |> parse
```
jparise: Assert that the parse tree is expected? Below, as well.

test/parser_test.exs
```diff
@@ -0,0 +1,554 @@
...
+      %Include{path: "foo.thrift"},
+      %Include{path: "bar.thrift"}
+    ]
+  end
+
+  test "parsing a byte constant" do
+    constant = "const i8 BYTE_CONST = 2;"
+    |> parse([:constants, :BYTE_CONST])
+
+
```
jparise: Extra newline

test/parser_test.exs
```diff
@@ -0,0 +1,554 @@
...
+  test "parsing a double constant" do
+    constant = "const double DOUBLE_CONST = 123.4"
+    |> parse([:constants, :DOUBLE_CONST])
+
+    assert constant == %Constant{name: :DOUBLE_CONST,
+                                 value: 123.4,
+                                 type: :double}
+  end
+
+  test "parsing a sting constant" do
```
jparise: "string"

test/parser_test.exs
```diff
@@ -0,0 +1,554 @@
...
+    }
+    """
+
+    expected_error = "Error: BadEx.evil, BadEx.bad share field number 1."
+    assert_raise RuntimeError, expected_error, fn ->
+      parse(program)
+    end
+
+  end
+  test "parsing a typedef" do
```
jparise: Missing newline

test/parser_test.exs
```diff
@@ -0,0 +1,554 @@
...
+    assert pinger.extends == :Pinger
+    assert other.extends == :Pinger
+  end
+
+  test "parsing a real thrift file" do
+    File.read!("./test/fixtures/app/thrift/tutorial.thrift")
+    |> parse
+
+    File.read!("./test/fixtures/app/thrift/shared.thrift")
+    |> parse
```
jparise: Assertion something about the resulting parse trees?

scohen: I'm just making sure they don't blow up...

lib/parser/models.ex
scohen: This prints too with the formatting we want. It also stops compilation, which I think is what we want to do.

src/thrift_parser.yrl
scohen: yes. IIRC Javascript was the first popular language to support this syntax.

test/parser_test.exs
scohen: Good call.

lib/parser/models.ex
```diff
@@ -0,0 +1,510 @@
+defmodule Thrift.Parser do
+
+  def parse(doc) do
```
jparise: Document the `parse/1` and `parse/2`, and consider moving them to their own file?

lib/parser/models.ex
```diff
@@ -0,0 +1,510 @@
...
+  defmodule Shell do
+    @yellow IO.ANSI.yellow
+    @red IO.ANSI.red
+    @reset IO.ANSI.reset
+
+    def warn(msg) do
+      IO.puts "#{@yellow}#{msg}#{@reset}"
+    end
+
+    def error(msg) do
```
jparise: This is unused, which I think is alright for now.

lib/parser/models.ex
```diff
@@ -0,0 +1,510 @@
...
+
+      @spec new(char_list, [%Function{}, ...], char_list) :: %Service{}
+      def new(name, functions, extends) do
+        %Service{name: atomify(name), extends: atomify(extends), functions: functions}
+      end
+    end
+
+    defmodule Schema do
+      @moduledoc """
+      A Thrift program.
```
jparise: s/program/schema/ ?

jparise:
Just added a few minor comments, but looks great overall!

lib/parser/models.ex
scohen: I can't move them to their own file, or I'll be redefining `Thrift.Parser`.

scohen:
@jparise We good to go?

PR #7, opened 2016-06-17
Simplify merge reduction work in Schema.new/2

jparise:
This approach introduces private merge/2 functions that match on the
individual model types that we want to merge into our schema. This
simplifies the top-level reduction.


scohen:
r+

lib/parser/models.ex
```diff
@@ -427,45 +427,52 @@
...
         |> Enum.reverse
-        |> Enum.reduce(
-          program,
-          fn
-            (const=%Constant{}, program) ->
-              %Schema{program | constants: Map.put(program.constants, const.name, const)}
+        |> Enum.reduce(schema, &merge(&2, &1))
+      end
+
+      @spec merge(t, header | definition) :: t
```
scohen: Can you make this a type too?

Model.t perhaps?

PR #11, opened 2016-06-21
Require at least Elixir 1.2

jparise:
With today's release of Elixir 1.3, now is a good time to drop compatibility
with much older releases. This also lets us use some of the nicer language
features introduced in Elixir 1.2, like the multi-alias syntax.


scohen:
R+

PR #14, opened 2016-06-23
Also test against OTP 19

jparise:


jparise:
Closing favor of #15.

PR #17, opened 2016-09-01
Support comment

fieldinrain:
T/O


jparise:
@fieldinrain Is this something you're planning on resubmitting later? Looks like some good stuff in there.

fieldinrain:
Hi, @jparise 

I am using elixir-thrift in my project to define the interface between server(elixir) and client(C#) .
And I extended it to support 

1. unsigned integer (u8/u16/u32/u64)
    the client side need some unsigned integer, but it is not the standard feature of Thrift

2. unix comment(# comment) 
    Output the comments written in thrift files, and we can write some annotation in comments to generate some special code (I made a source code generator.)

I wonder if it can be used for the general usage, especially unsigned integer. So I close the PR and merged it to my own master branch. 
If you think that my PR can be accepted, I will resubmit it later.

Best regards

fieldinrain

jparise:
Got it, and I'm glad you're finding our library useful!

I think some support for (2) would make sense in the main elixir-thrift package, but (1) probably isn't a good fit because it's non-standard.

PR #22, opened 2016-10-02
Full test coverage for Mix.Tasks.Compile.Thrift

jparise:
Add test coverage for the last two remaining error cases. This brings
this module's unit test coverage up to 100%.

This also includes one functional change: we now redirect :stderr output
to :stdout so we can more consistently capture and display output from
the Thrift compiler executable.


jparise:
@scohen, this look good to you?

scohen:
r+

Issue #16, opened 2016-07-18
Consider Docker-based Thrift build


jparise:
The bulk of our CI (Travis) build time is spent (re)building the thrift compiler tool. (There unfortunately isn't a readily available Thrift 0.9.3 .deb package.)

Consider using a Docker-based build package like in dantswain/thrash#8.

dantswain:
If you're interested, I could probably make a PR for this.

jparise:
That would be great, @dantswain!

jparise:
Closed by #24

PR #24, opened 2016-10-16
Use docker to run thrift for travis

dantswain:
Re #16 

The basic gist is:
* `.thrift-docker` wraps thrift in a `docker run` call.
* When travis runs, it `mv .thrift-docker thrift` and sets the PATH so that the wrapper shows up when we just run `thrift`.

The main caveat here is that the wrapper sets the working directory to the fixture app's root directory.  There might be a more flexible way to do this - it just gets a little tricky with paths.


.travis.yml
```diff
@@ -6,27 +6,14 @@ otp_release:
...
-    - flex
-    - g++
-    - libboost-dev
-    - libboost-test-dev
-    - libboost-program-options-dev
-    - libevent-dev
-    - libtool
-    - libssl-dev
-    - pkg-config
+sudo: required
```
jparise: Is `sudo` required for docker?

dantswain: In general, it depends on how you have docker set up.  Travis requires it: https://docs.travis-ci.com/user/docker/

dantswain:
> 1. Would it be cleaner to use Travis's environmental variables to determine paths?

I don't _think_ that helps here. The path trickiness that I'm referencing has more to do with setting the working directory when you run docker.  We have to have the wrapper script in the PATH (which is easy enough to modify in .travis.yml) so that we can execute `System.cmd("thrift", _)` without any extra config.  Then when the wrapper executes `docker run`, it sets the working directory so that it's in a spot where the arguments make sense wrt the mounted volume.

> 2. Could we use the existing Makefile instead of the hidden .thrift-docker wrapper? If not, we should remove the Makefile because it's only used by the CI environment.

I'd say it's cleanest just to remove the Makefile.  I could move the 'hidden' file to another directory and have it not be a hidden file if you prefer.

FWIW when I did this for thrash, I set it up so that the thrift executable could be specified via an environment variable.  Then instead of the wrapper script I just do `THRIFT="docker run .... etc" mix test`.

jparise:
> I'd say it's cleanest just to remove the Makefile. I could move the 'hidden' file to another directory and have it not be a hidden file if you prefer.

Yes, I think it makes sense to remove the Makefile. And I'd prefer not to have the hidden file in the root directory. We could add a top-level `ci/` directory that contains "helper" files for Travis. I've seen that approach used in other projects.

> FWIW when I did this for thrash, I set it up so that the thrift executable could be specified via an environment variable. Then instead of the wrapper script I just do THRIFT="docker run .... etc" mix test.

I see. You're checking for the environment variable in your Mix task. Have you found that useful beyond this case?

jparise:
Actually, I wonder if there's a nice way to use an alternate Mix configuration profile so that we could use the existing `:thrift_executable` task option.

dantswain:
(On my phone, probably won't get formatting right)

You could definitely check for the CI env var in the config.exs file and set the executable config based on that. 

I've used the env var approach a few places. It's kind of inspired by the way a lot of config/Make tools work where you might set `CC=/opt/gcc-some-version/bin/gcc` etc. It's not necessarily the _right_ way but it works.  If I wanted to make a principled argument I could say that the path of an executable is a property of the environment rather than of the project, but I'd be lying if I said I did it that way from the start ;)

jparise:
#26 is a compromise that adds support for the `THRIFT` environment variable for the fixture-based test environment. I think that should be sufficient to simplify the use of the Docker-based Thrift package proposed here.

I'm not opposed to adding first-class support for `THRIFT` to the main Mix task, but it's also easy for people to add something similar to their `mix.exs` scripts. I think this approach composes pretty well.

jparise:
This is great! Thanks, @dantswain!

PR #31, opened 2016-11-09
File Parser / Resolver

scohen:
The parser can now take a single thrift file and follow all its
includes and parse them as well. The output is stored in a FileGroup,
which is returned.

Added a resolver to handle resolution of struct references. As
references are added to the FileGroup, they are placed in the resolver
and after parsing is complete, added to the resolutions field in the
FileGroup.

@pguillory @jparise

### DIFF
--- a/lib/parser/file_group.ex
+++ b/lib/parser/file_group.ex
@@ -12,8 +12,6 @@ defmodule Thrift.Parser.FileGroup do
     Schema,
   }
 
-  defdelegate fix_refs(), to: Resolver
-
   @type t :: %FileGroup{
     parsed_files: %{FileRef.thrift_include => %ParsedFile{}},
     schemas: %{FileRef.thrift_include => %Schema{}}}
--- a/lib/parser/file_group.ex
+++ b/lib/parser/file_group.ex
@@ -1,4 +1,10 @@
 defmodule Thrift.Parser.FileGroup do
+  @moduledoc """
+  Represents a group of parsed files. When you parse a file, it might include other thrift files.
+  These files are in turn accumulated and parsed and added to this module.
+  Additionally, this module allows resolution of the names of Structs / Enums / Unions etc across
+  files.
+  """
   alias Thrift.Parser.{
     FileGroup,
     FileRef,
--- a/lib/parser/file_group.ex
+++ b/lib/parser/file_group.ex
@@ -1,0 +1,63 @@
+defmodule Thrift.Parser.FileGroup do
+  alias Thrift.Parser.{
+    FileGroup,
+    FileRef,
+    Resolver,
+    ParsedFile
+  }
+
+  alias Thrift.Parser.Models.{
+    Field,
+    StructRef,
+    Schema,
+  }
+
+  defdelegate fix_refs(), to: Resolver
+
+  @type t :: %FileGroup{
+    parsed_files: %{FileRef.thrift_include => %ParsedFile{}},
+    schemas: %{FileRef.thrift_include => %Schema{}}}
+
+  defstruct parsed_files: %{}, schemas: %{}, resolutions: %{}
+
+  def add(file_group, parsed_file) do
+    file_group = add_includes(file_group, parsed_file)
+    new_parsed_files = Map.put(file_group.parsed_files, parsed_file.name, parsed_file)
+    new_schemas = Map.put(file_group.schemas, parsed_file.name, parsed_file.schema)
+
+    Resolver.add(parsed_file)
+    %__MODULE__{file_group |
+                parsed_files: new_parsed_files,
+                schemas: new_schemas}
+  end
+
+  def add_includes(group=%__MODULE__{},
+                   %ParsedFile{schema: schema, file_ref: file_ref}) do
+
+    Enum.reduce(schema.includes, group, fn(include, file_group) ->
+      parsed_file = file_ref.path
+      |> Path.dirname
+      |> Path.join(include.path)
+      |> FileRef.new
+      |> ParsedFile.new
+      add(file_group, parsed_file)
+    end)
+  end
+
+  def resolve(%FileGroup{}=group, %Field{type: %StructRef{}=ref}=field) do
+    %Field{field | type: resolve(group, ref)}
+  end
+
+  def resolve(%FileGroup{resolutions: resolutions}, %StructRef{referenced_type: type_name}) do
+    resolutions[type_name]
+  end
+
+  def resolve(%FileGroup{resolutions: resolutions}, path) when is_atom(path) do
+    resolutions[path]
+  end
+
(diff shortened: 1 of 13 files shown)


lib/parser/models.ex
```diff
@@ -423,14 +423,15 @@ end
       Services hold RPC functions and can extend other services.
       """
 
-      @type t :: %Service{name: String.t, extends: String.t, functions: [%Function{}]}
-      defstruct name: nil, extends: nil, functions: []
+      @type t :: %Service{name: String.t, extends: String.t, functions: %{atom => %Function{}}}
+      defstruct name: nil, extends: nil, functions: {}
```
jparise: Should this be `%{}`?

lib/parser/file_ref.ex
```diff
@@ -0,0 +1,18 @@
+defmodule Thrift.Parser.FileRef do
+  @type thrift_include :: String.t
+  @type t :: %__MODULE__{path: String.t, include_name: String.t, contents: String.t}
+  defstruct path: nil, include_name: nil, contents: nil
+
+  def new(path) do
+    thrift_file = """
```
jparise: Will this approach result in confusing line numbers for parse errors?

lib/parser/resolver.ex
```diff
@@ -0,0 +1,66 @@
...
+
+  def stop do
+    Agent.stop(__MODULE__)
+  end
+
+  def print do
+    Agent.get(__MODULE__, &IO.inspect(&1))
+  end
+
+  def add(f=%ParsedFile{}) do
```
jparise: I see you somethings write `x=%{...}` and sometimes `%{...}=x`. We should use a consistent style.

lib/parser/file_ref.ex
scohen: I'm open to suggestions. IIRC this was your idea. 

put it at the end of the file?

lib/parser/resolver.ex
scohen: I used to write it x=%{} but I think the convention now is %{}=x. I'll fix.

lib/parser/file_group.ex
```diff
@@ -0,0 +1,63 @@
...
+    ParsedFile
+  }
+
+  alias Thrift.Parser.Models.{
+    Field,
+    StructRef,
+    Schema,
+  }
+
+  defdelegate fix_refs(), to: Resolver
```
pguillory: What is this used for?

scohen: deleted.

lib/parser/file_ref.ex
```diff
@@ -0,0 +1,14 @@
+defmodule Thrift.Parser.FileRef do
+  @type thrift_include :: String.t
+  @type t :: %__MODULE__{path: String.t, include_name: String.t, contents: String.t}
+  defstruct path: nil, include_name: nil, contents: nil
+
+  def new(path) do
+    thrift_file = File.read!(path) <> "\n__file__ \"#{path}\""
```
pguillory: I don't really understand the __file__ aspect of this. Where does it get used?

scohen: It gets used when we're building the names of things. The thrift module is the same as the file name, so we need the filename somewhere. However, leex and yecc deal with byte streams, and there's no way to get the file. 
This is a hack around that.

lib/parser/file_group.ex
```diff
@@ -0,0 +1,61 @@
+defmodule Thrift.Parser.FileGroup do
```
jparise: `@moduledoc` would be a useful way to introduce this concept.

lib/parser/resolver.ex
```diff
@@ -0,0 +1,36 @@
+defmodule Thrift.Parser.Resolver do
```
jparise: A little `@moduledoc` here too please

lib/parser/parser.ex
```diff
@@ -45,4 +45,24 @@ defmodule Thrift.Parser do
         Map.get(next, part)
     end)
   end
+
+  @doc """
+  Takes a directory that contains thrift files and a list
```
jparise: Why is this called `parse_file` if it works with directories?

lib/parser/file_ref.ex
jparise: Probably deserves a comment though.

lib/parser/models.ex
```diff
@@ -509,28 +526,36 @@ end
...
       defp merge(schema, %Service{} = service) do
-        %Schema{schema | services: Map.put(schema.services, service.name, service)}
+        %Schema{schema | services: Map.put(schema.services, service.name, update_name(schema, service))}
       end
 
       defp merge(schema, {:typedef, actual_type, type_alias}) do
         %Schema{schema | typedefs: Map.put(schema.typedefs, atomify(type_alias), actual_type)}
       end
+
+      defp update_name(%{module: nil}, model) do
```
jparise: Can we think of a better name for this function? Looking at how it's called above, the idea of "name" is non-obvious and ambiguous.

test/test_helper.exs
```diff
@@ -1 +1,70 @@
...
+
+    file_path
+    |> Path.dirname
+    |> File.mkdir_p!
+
+    File.write!(file_path, contents)
+    file_relative_path
+  end
+
+  def tmp_dir do
```
jparise: Could we use `System.tmp_dir!` instead?

lib/parser/parser.ex
scohen: Because the docs are wrong.

test/test_helper.exs
scohen: that doesn't reliable create a temporary directory though. The idea behind this is that each test makes its own directory and puts thrift files into it. That makes it easy to clean up.

lib/parser/file_ref.ex
```diff
@@ -0,0 +1,17 @@
+defmodule Thrift.Parser.FileRef do
+  @type thrift_include :: String.t
+  @type t :: %__MODULE__{path: String.t, include_name: String.t, contents: String.t}
+  defstruct path: nil, include_name: nil, contents: nil
+
+  def new(path) do
+    # We include the __file__ here to hack around the fact that leex and yeec don't
```
jparise: `s/yeec/yecc/`

test/test_helper.exs
jparise: I see. The thing that I don't like about this particular approach is that it relies on `mktemp(1)`, which limits the portability of the tests a small amount. It would be better to do this with built-in functions.

Perhaps something more like:

```elixir
tmp_path = Path.join(System.tmp_dir!(), Integer.to_string(System.unique_integer))
System.mkdir tmp_path
```

... or something like that.

PR #32, opened 2016-11-10
Add model generation

pguillory:
Run 'mix generate.thrift' to generate Elixir models corresponding to Thrift
structs/enums/etc.


lib/mix/tasks/generate.thrift.ex
```diff
@@ -0,0 +1,28 @@
+defmodule Mix.Tasks.Generate.Thrift do
```
jparise: Let's add some `@moduledoc` that documents the task and its options.

lib/generator/models.ex
```diff
@@ -0,0 +1,175 @@
...
+  end
+
+  def namespace(%{namespaces: namespaces}) do
+    case namespaces do
+      %{ex: %{path: path}} ->
+        path |> String.split(".") |> Enum.map(&Macro.camelize/1) |> Enum.join(".")
+      %{py: %{path: path}} ->
+        path |> String.split(".") |> Enum.map(&Macro.camelize/1) |> Enum.join(".")
+      %{} ->
+        raise "Schema has no namespace"
```
scohen: Schemas don't need a namespace.

lib/mix/tasks/generate.thrift.ex
```diff
@@ -0,0 +1,28 @@
...
+
+    thrift_dir = Keyword.get(opts, :thrift_dir, "thrift")
+    lib_dir = Keyword.get(opts, :lib_dir, "lib")
+
+    for thrift_file <- list_thrift_files(thrift_dir) do
+      thrift_file
+      |> Thrift.Generator.Models.generate
+      |> Enum.each(fn {model_file, source} ->
+        path = Path.join(lib_dir, model_file)
+        :ok = path |> Path.dirname |> File.mkdir_p!
```
jparise: Do we need to match against `:ok` when `File.mkdir_p!` raises an exception on failure?

lib/generator/models.ex
```diff
@@ -0,0 +1,175 @@
...
+      |> Kernel.<>(".ex")
+      source = Macro.to_string(quoted)
+      {filename, source}
+    end)
+  end
+
+  def namespace(%{namespaces: namespaces}) do
+    case namespaces do
+      %{ex: %{path: path}} ->
+        path |> String.split(".") |> Enum.map(&Macro.camelize/1) |> Enum.join(".")
```
scohen: What's going on here? Why do you need to modify the passed in namespace?

lib/mix/tasks/generate.thrift.ex
```diff
@@ -0,0 +1,28 @@
...
+        :ok = path |> File.write!(source)
+        Mix.shell.info "Generated #{path}"
+      end)
+    end
+  end
+
+  defp list_thrift_files(dir) do
```
jparise: I believe this is basically the same as `Mix.Utils.extract_files([dir], [:thrift])`.

lib/generator/models.ex
```diff
@@ -0,0 +1,175 @@
...
+    end)
+    quote do
+      defmodule unquote(name) do
+        @moduledoc unquote("Auto-generated Thrift exception #{inspect name}")
+        defstruct unquote(parts)
+      end
+    end
+  end
+
+  # Zero values for base types
```
scohen: I usually call these default values.

test/fixtures/app/thrift/shared.thrift
```diff
@@ -29,6 +29,7 @@ namespace java shared
 namespace perl shared
 namespace php shared
 namespace haxe shared
+namespace ex shared
```
jparise: I think I'd prefer `elixir` spelled out. `ex` is too ambiguous, especially when that's not the name of the language.

lib/generator/models.ex
```diff
@@ -0,0 +1,175 @@
...
+      source = Macro.to_string(quoted)
+      {filename, source}
+    end)
+  end
+
+  def namespace(%{namespaces: namespaces}) do
+    case namespaces do
+      %{ex: %{path: path}} ->
+        path |> String.split(".") |> Enum.map(&Macro.camelize/1) |> Enum.join(".")
+      %{py: %{path: path}} ->
```
jparise: Why are we handling Python namespaces?

lib/generator/models.ex
```diff
@@ -0,0 +1,175 @@
...
+
+  defp load_refs(path) when is_binary(path) do
+    load_refs([path], %{})
+  end
+
+  defp load_refs([], refs) do
+    refs
+  end
+
+  defp load_refs([path | paths], refs) do
```
scohen: my PR does this for you.

test/generator/models_test.exs
```diff
@@ -0,0 +1,96 @@
...
+        3: optional i32 num2 = 5;
+        4: optional bool b1;
+        # Not yet supported.
+        # 5: optional bool b2 = true;
+      }
+      """
+    files = thrift |> parse |> generate
+    assert [{"my_struct.ex", contents}] = files
+    assert contents == String.trim """
+      defmodule(MyStruct) do
```
scohen: I don't know if we should do this now or soon, but
* Make a type spec for the struct. 
* Make a new function that enforces values / types and takes opts
* Make a type spec for the options.

lib/generator/models.ex
```diff
@@ -0,0 +1,175 @@
...
+        end
+      Map.has_key?(schema.refs, type) ->
+        {schema, name, _value} = schema.refs[type]
+        zero(schema, %{referenced_type: name})
+      true ->
+        raise "Unknown type: #{inspect type}"
+    end
+  end
+
+  # def zero(_schema, other) do
```
jparise: Remove?

lib/generator/models.ex
```diff
@@ -0,0 +1,175 @@
...
+    quote do
+      defmodule unquote(name) do
+        @moduledoc unquote("Auto-generated Thrift exception #{inspect name}")
+        defstruct unquote(parts)
+      end
+    end
+  end
+
+  # Zero values for base types
+  def zero(_schema, :bool), do: false
```
jparise: Could these be private?

.gitignore
```diff
@@ -1,6 +1,7 @@
 /_build
 /deps
 /doc
+/lib/generated
```
jparise: Can you say more about why this is here for this library package? This might be useful for applications that use this package, but I'm not sure how it works in this context.

lib/mix/tasks/generate.thrift.ex
pguillory: Handy.

lib/generator/models.ex
pguillory: It was a fallback to be able to scan our existing schemas repo without adding an "elixir" namespace to each one. I'll take it out.

test/fixtures/app/thrift/shared.thrift
pguillory: Done.

scohen:
@pguillory Can we get the coverage back up to 99.3%?

jparise:
I'd also like to avoid such a large test coverage decrease before this is merged.

lib/mix/tasks/thrift.generate.ex
```diff
@@ -0,0 +1,30 @@
...
+    * `--lib-dir` - Directory under which to place generated .ex files. (Default: ./lib)
+  """
+
+  def run(args) do
+    {opts, _, _} = OptionParser.parse(args)
+
+    thrift_dir = Keyword.get(opts, :thrift_dir, "thrift")
+    lib_dir = Keyword.get(opts, :lib_dir, "lib")
+
+    for thrift_file <- Mix.Utils.extract_files([thrift_dir], "*.thrift") do
```
jparise: You can just pass `"thrift"` as the argument. `extract_files` can take a list of file extensions or a glob pattern.

lib/generator/models.ex
pguillory: The namespace determines both the name of the generated module and the filename in which it is written. By convention Whatever.MyModule goes in whatever/my_module.ex. (http://elixir-lang.org/docs/master/elixir/naming-conventions) This does the conversion from the underscore filename version to the CamelCase module name version. It seems like the namespaces for other languages tend to be defined in the underscore style, so it would be convenient to use that for Elixir also.

lib/generator/models.ex
pguillory: Made it default to nil, which will put things in the root namespace.

lib/generator/models.ex
pguillory: I might have picked it up from Go. (https://golang.org/ref/spec#The_zero_value) This is a bit distinct from the default value, because that's a thing you can define or not. This is the "default default" if you don't define a default.

.gitignore
pguillory: During development I've been running the mix task a lot against different inputs and looking at the generated code. It was a bit distracting having to selectively `git add` certain things. I don't think this will affect users of the library.

test/generator/models_test.exs
pguillory: As in a setter function?

lib/mix/tasks/thrift.generate.ex
pguillory: I tried passing "thrift" or ".thrift" but it returned an empty list.

PR #35, opened 2016-11-15
Binary protocol

scohen:
This is a big change, but everything in /test/support is autogenerated. 

Also, the first commit has already been reviewed.
@jon @pguillory


preston:
@scohen Here I thought I was going senile... but I think you have the wrong "Preston". 😀

lib/generator/models.ex
```diff
@@ -0,0 +1,172 @@
...
+  defp zero(_schema, :bool), do: false
+  defp zero(_schema, :byte), do: 0
+  defp zero(_schema, :i8), do: 0
+  defp zero(_schema, :i16), do: 0
+  defp zero(_schema, :i32), do: 0
+  defp zero(_schema, :i64), do: 0
+  defp zero(_schema, :double), do: 0.0
+  defp zero(_schema, :string), do: ""
+  defp zero(_schema, :binary), do: ""
+  defp zero(_schema, :slist), do: ""
```
jparise: Are we going to support `slist`? If we do, we should also include `senum`. I'm fine leaving them both out because they've been deprecated for a long time: https://thrift.apache.org/docs/idl#senum

lib/parser/parser.ex
```diff
@@ -52,17 +52,14 @@ defmodule Thrift.Parser do
...
-    file_group = %FileGroup{}
+    file_group = FileGroup.new(file_path)
     |> FileGroup.add(parsed_file)
 
-    resolutions = Resolver.get()
-    Resolver.stop()
-
-    %{file_group | resolutions: resolutions}
+    file_group = FileGroup.update_resolutions(file_group)
+    file_group
```
jparise: Or just `FileGroup.update_resolutions(file_group)`

lib/protocols/binary.ex
```diff
@@ -0,0 +1,212 @@
+defmodule Thrift.Protocols.Binary do
+  alias Thrift.Parser.FileGroup
+
+  # field types
```
jparise: Mention these are the official field type IDs from the Thrift spec?

lib/protocols/binary.ex
```diff
@@ -0,0 +1,212 @@
...
+          }
+
+  alias Thrift.Parser.Models.{
+    Exception,
+    Field,
+    Struct,
+    TEnum,
+  }
+
+  def bool_to_int(false), do: 0
```
jparise: Make these private and move them below (like `to_mesage_type`)?

mix.exs
```diff
@@ -36,12 +37,29 @@ defmodule Thrift.Mixfile do
      []
   end
 
+  def erlc_paths(:prod) do
```
jparise: These can all be private.

test/fixtures/app/thrift/simple.thrift
```diff
@@ -0,0 +1,31 @@
...
+  5: i16 nineties_era_color,
+  6: double mint_gum,
+  7: string username,
+  8: list<User> friends,
+  9: map<i8, string> my_map,
+  10: set<i32> blocked_user_ids,
+  11: optional list<i32> optional_integers,
+}
+
+/*
```
jparise: Can we remove the commented code?

test/fixtures/app/thrift/tutorial.thrift
```diff
@@ -106,7 +107,7 @@ enum Operation {
 struct Work {
   1: i32 num1 = 0,
   2: i32 num2,
-  3: Operation op,
+  #3: Operation op,
```
jparise: Does this no longer work?

lib/generator/models.ex
scohen: I'd leave them out, since they're deprecated. This change was from the first commit. If it's alright, I'll edit in a subsequent PR.

test/fixtures/app/thrift/tutorial.thrift
scohen: I'm not sure @pguillory commented that out.

dantswain:
There isn't much going on here by way of documentation (or with the previous PR).  Is there a plan for that?  IIUC this is a "next release" development step so I could understand not updating the main README.  OTOH it would be easier to follow along with the PRs if there was some documentation.  Maybe a "README.next_release.md" or something?

bench/binary_protocol.exs
```diff
@@ -0,0 +1,56 @@
...
+  @thrift_file_path "./test/fixtures/app/thrift/simple.thrift"
+  alias Thrift.Parser
+  alias Thrift.Protocols.Binary
+  import ParserUtils
+
+  setup_all do
+    file_group = parse_thrift(@thrift_file_path)
+    |> compile_module
+
+    {:ok, :ok}
```
jparise: Looking at [the docs](http://elixir-lang.org/docs/stable/ex_unit/ExUnit.Callbacks.html#module-context), it appears you can return just `:ok` if you don't have any metadata to return.

lib/parser/file_group.ex
```diff
@@ -48,6 +55,27 @@ defmodule Thrift.Parser.FileGroup do
     end)
   end
 
+  def update_resolutions(file_group) do
+    # since in a file, we can refer to things defined in that file in a non-qualified
```
jparise: Missing a word?

test/generator/models_test.exs
```diff
@@ -0,0 +1,263 @@
...
+  use ExUnit.Case, async: true
+
+  import Thrift.Generator.Models
+
+  setup %{test: test} do
+    dir = Path.join([System.tmp_dir!, to_string(__MODULE__), to_string(test)])
+    File.rm_rf!(dir)
+    File.mkdir_p!(dir)
+    on_exit fn ->
+      # IO.inspect dir
```
jparise: Remove?

scohen:
@dantswain Right now, we're targeting a branch that we'll merge into master when we have full serialization / deserialization of thrift types. We'll add docs toward the end; we're still experimenting with a lot (see the TODOs). I don't think it makes sense to document when we still have so much stuff left up in the air.

bench/binary_protocol.exs
scohen: @jparise this isn't ExUnit, it's Benchfella. 
Returning :ok, yields
```
Settings:
  duration:      1.0 s

## BinaryProtocolBenchmark
** (RuntimeError) Expected BinaryProtocolBenchmark.setup_all/0 to return {:ok, <term>}. Got :ok
Skipping all tests in BinaryProtocolBenchmark


Finished in 0.14 seconds
```

lib/parser/file_group.ex
scohen: I don't think so... 
"Since in a file, we can refer to things defined in that file in a non-qualified way..." makes sense to me.

scohen:
@jparise Having the files in ext as a git submodule made it extremely difficult to configure mix to include them in the project while under test; It appears you only get one 'shot' at an include directory, and we'd have two. "src and ext/thrift/lib/erl/include". 

I know it's uglier, but it's easier for people to get started and just works.

dantswain:
@scohen Won't the ext directory go away eventually anyways if this is a 100% Elixir implementation?

scohen:
@dantswain Yes.

jparise:
@scohen I think I'm just missing the purpose of adding those Erlang runtime files a second time because the ones under `ext/` were already being compiled and "linked" correctly before.

In other words, it's not specifically the `git submodule` bit that I'm advocating but rather why we should have them in two places.

PR #40, opened 2016-11-17
Module resolver

scohen:
Implemented a resolver that resolves thrift modules into their destination namespaces. Applied it to serializer. 

Now nested structs work even across modules and namespaces. 

@jparise @pguillory


lib/parser/file_group.ex
```diff
@@ -92,4 +99,35 @@ defmodule Thrift.Parser.FileGroup do
...
+    |> Enum.map(&String.to_atom/1)
+
+    case file_group.ns_mappings[thrift_module] do
+      nil ->
+        Module.concat(Elixir, struct_name)
+      namespace = %Namespace{} ->
+        namespace.path
+        |> String.split(".")
+        |> Enum.map(&Macro.camelize/1)
+        |> Enum.join(".")
```
pguillory: There's duplication between this code and Thrift.Generator.Models.namespace that we should consolidate.

lib/protocols/binary.ex
```diff
@@ -87,56 +101,53 @@ defmodule Thrift.Protocols.Binary do
...
-      defp to_message_type(:oneway), do: 4
     end
   end
 
-  def build(file_group, dest_module) do
-    schema = file_group.schemas["simple"]
-    user_struct = schema.structs[:User]
-    generate_serializer(dest_module, file_group, user_struct)
+  def build(file_group, struct) do
+    alias Thrift.Generator.Models.BinaryProtocol, as: Deserializer
```
pguillory: This ought to go up at the top.

pguillory: (nit)

lib/protocols/binary.ex
```diff
@@ -186,8 +193,11 @@ defmodule Thrift.Protocols.Binary do
                   {composite, _} ->
                     composite
 
-                    %TEnum{} ->
+                  %TEnum{} ->
                     :i8
```
pguillory: Enums are encoded as an i32.

test/support/lib/parser_utils.ex
```diff
@@ -1,39 +1,42 @@
...
+  defstruct is_evil: false, user_id: 0, number_of_hairs_on_head: 0, amount_of_red: 0, nineties_era_color: 0, mint_gum: 0.0, username: "", friends: [], my_map: %{}, blocked_user_ids: MapSet.new(), optional_integers: []
+end
+
+defmodule Nesting do
+  defstruct user: nil, nested: nil
+end
+
+defmodule Shared.SharedStruct do
+  defstruct key: nil, value: nil
+end
```
pguillory: These are tightly coupled to specific test fixtures. It would be nice if we could use the code generator to get them dynamically.

scohen: I'm open to suggestions. 
The problem here is that we need to use the modules in the functions below. I've tried hooking into __before_compile__ but no luck.

lib/parser/file_group.ex
scohen: I didn't want to consolidate until I was sure you're done in there.

bench/binary_protocol.exs
```diff
@@ -32,14 +32,23 @@ defmodule BinaryProtocolBenchmark do
     {:ok, elixir_users: elixir_users, erlang_users: erlang_users}
   end
 
-  bench "erlang serialization" do
+  bench "erlang serialization (converted to binary)" do
     for user <- bench_context[:erlang_users] do
       serialize_user(user)
```
jparise: Can you add `convert_to_binary: true` explicitly here? I think that would make this even clearer.

lib/generator/models.ex
```diff
@@ -12,24 +12,41 @@ defmodule Thrift.Generator.Models do
...
       schema
       |> Map.put(:file_group, file_group)
-      |> generate_schema(output_dir)
+      |> generate_schema
+      |> write_schema_to_file(output_dir)
     end)
   end
 
-  def generate_schema(schema, output_dir) do
+  def generate_to_string!(%FileGroup{}=file_group) do
```
jparise: Would it be too tricky to express this as an implementation of the [`String.Chars` protocol](http://elixir-lang.org/docs/stable/elixir/String.Chars.html)?

test/support/lib/parser_utils.ex
```diff
@@ -1,39 +1,42 @@
...
-        defmodule BinaryProtocol do
-          unquote(Binary.primitive_serializers)
-          unquote(Binary.build(file_group, Testing.Simple.User))
-        end
-      end
-    end
-    #    quoted |> Macro.to_string |> IO.puts
-    quoted |> Code.compile_quoted
+    Thrift.Generator.Models.generate_to_string!(file_group)
+#    |> print_compiled_code
```
jparise: Remove?

lib/generator/models.ex
```diff
@@ -12,24 +12,41 @@ defmodule Thrift.Generator.Models do
...
+    Enum.flat_map(file_group.schemas, fn {_, schema} ->
+      schema
+      |> Map.put(:file_group, file_group)
+      |> generate_schema
+    end)
+    |> Enum.reverse
+    |> Enum.map(fn {_, code} ->
+      Macro.to_string(code)
+    end)
+    |> Enum.join("\n")
```
pguillory: What's up with the join? Previously a single .thrift input file would generate multiple output files, one for each struct/enum/etc. Does this change that?

lib/generator/models.ex
scohen: I'd worry about doing that, since stringifying FileGroup anywhere would expand it into code. I'd rather not do that when printing out state.

PR #41, opened 2016-11-17
Add deserialization tests

pguillory:


lib/generator/binary_protocol.ex
```diff
@@ -251,7 +251,7 @@ defmodule Thrift.Generator.Models.BinaryProtocol do
 
   def map_value_deserializer(:string, name, _file_group) do
     quote do
-      defp unquote(name)(<<string_size::size(64), rest::binary>>, key, [map | stack], remaining) do
+      defp unquote(name)(<<string_size::32-signed, rest::binary>>, key, [map | stack], remaining) do
```
jparise: Just want to verify that you intended this to go from 64- to 32-bits (i.e. this is a bug fix).

PR #42, opened 2016-11-17
Add enum 'valid?' method

pguillory:
This generates a valid?/1 method on enum models. It tests whether a given
number is a valid ordinal for the given enum.


jparise:
Do you think `member?/1` might be a better name? That would match `Enum.member?/1`.

Also, could you include tests for this?

PR #43, opened 2016-11-18
Fixed parsing errors

scohen:
@pguillory @jparise @dantswain


lib/parser/models.ex
```diff
@@ -108,6 +108,11 @@ end
       nil
     end
 
+    def cast(type, %{}=val) do
+      # this is for TEnumValues
```
pguillory: Can it match on `%TEnum{}` then, instead of `%{}`?

lib/parser/models.ex
```diff
@@ -221,6 +232,29 @@ end
...
 
+    defmodule TEnumValue do
+      @moduledoc """
+      A reference to an enum value
+      For example, in a constant or default value.
+
+         const string DEFAULT_WEATHER = Weather.SUNNY;
+     """
+      @type t :: %TEnumValue{enum_name: atom, enum_value: atom, type: atom}
+      defstruct enum_name: nil, enum_value: nil, type: nil
```
pguillory: Nearby structs have fields like `name` and `value` or `values`. Would that make sense here, for consistency, rather than `enum_name` and `enum_value`?

test/parser_test.exs
```diff
@@ -174,6 +175,132 @@ defmodule ParserTest do
...
+    |> parse([:constants, :clothes_to_weather])
+
+    assert constant == %Constant{
+      name: :clothes_to_weather,
+      type: {:map, {:string, %StructRef{referenced_type: :Weather}}},
+      value: %{
+        "gloves" => %TEnumValue{
+        enum_name: :Weather,
+        enum_value: :SNOWY,
+        type: %StructRef{referenced_type: :Weather}},
```
pguillory: Indentation is off here.

test/parser_test.exs
```diff
@@ -174,6 +175,132 @@ defmodule ParserTest do
...
+
+        %TEnumValue{
+          enum_name: :Weather,
+          enum_value: :SUNNY,
+          type: %StructRef{referenced_type: :Weather}} => "Yay, it's sunny!"}}
+  end
+
+  test "parsing a map constant with enum values as values" do
+    constant = """
+    const map<string, Weather> clothes_to_weather = {
```
pguillory: Briefly mis-read it as "clothes_to_wear", which I kind of like. :)

lib/parser/models.ex
scohen: That's not defined yet, so sadly, no.

lib/parser/models.ex
scohen: I did this intentionally; in the binary protocol we make a lot of assumptions about what name and value mean, an din this case, we can't just do the generic thing (resolve the struct name) and replace the original with the resolution. I'm trying to be clear that we're talking about a window into an enum.

PR #44, opened 2016-11-18
Add ThriftTestCase

pguillory:


test/generator/models_test.exs
```diff
@@ -1,263 +1,122 @@
...
-        end
-        defmacro(evil) do
-          32
-        end
-        # ...
-      end
-      """
+  use ThriftTestCase, cleanup: false
+
+  @thrift_file name: "enums.thrift", contents: """
```
scohen: Are the filenames getting us anything?

test/generator/models_test.exs
```diff
@@ -1,263 +1,122 @@
...
-        # ...
-      end
-      """
+  @thrift_file name: "exceptions.thrift", contents: """
+  exception ApplicationException {
+    1: string message,
+    2: required i32 count,
+    3: optional string reason
+    optional string other;
+    optional string fixed = "foo"
```
scohen: FYI, unnumbered fields are a bad thing in practice. I think we wan on this.

test/support/lib/thrift_test_case.ex
```diff
@@ -0,0 +1,96 @@
...
+      end
+    end)
+
+    quote do
+      unquote_splicing(Enum.map(modules, fn module ->
+        quote do: alias unquote(module)
+      end))
+      unquote_splicing(Enum.map(modules, fn module ->
+        quote do: require unquote(module)
+      end))
```
scohen: can these two be combined?
```elixir
{aliases, requires} = modules
|> Enum.reduce({[], []}, fn module, {aliases, requires} -> 
 {[quote(do: alias unquote(module)) | aliases],
  [quote(do: require unquote(module)) | requires]
 }
end)

quote do
   unquote_splicing(aliases)
   unquote_splicing(requires)
   ...
end

File: test/support/lib/thrift_test_case.ex [hunk 4]
```diff
@@ -0,0 +1,96 @@
...
+
+      setup_all do
+        on_exit fn ->
+          unquote(if Keyword.get(opts, :cleanup, true) do
+            quote do: File.rm_rf!(unquote(dir))
+          else
+            quote do: IO.puts IO.ANSI.format([:yellow, unquote("Leaving files in #{inspect dir}")])
```
scohen: classy!

test/generator/models_test.exs
```diff
@@ -1,263 +1,122 @@
...
-        defmacro(banned) do
-          6
-        end
-        defmacro(evil) do
-          32
-        end
-        # ...
-      end
-      """
+  use ThriftTestCase, cleanup: false
```
scohen: cleanup: true

test/support/lib/thrift_test_case.ex
pguillory: The code seems simpler as it is. What's the advantage to doing it the other way?

test/generator/models_test.exs
pguillory: Fixed.

test/support/lib/thrift_test_case.ex
scohen: i guess simple is in the eye of the beholder.

scohen: Also, when generating macros, I like to keep my unquotes as simple function calls.

PR #45, opened 2016-11-21
Added full thrift testing framework

scohen:
Thrift tests can now create erlang files using the thrift compiler.
When this happens, records are generated in special modules as are
conversion functions to facilitate testing.

This uncovered bugs with the deserializer, which I commented out until
they're fixed. It also revealed bugs with serializing container types,
which will be fixed in a subsequent PR.

@jparise @pguillory @dantswain

### DIFF
--- a/lib/parser/file_group.ex
+++ b/lib/parser/file_group.ex
@@ -86,19 +86,31 @@ defmodule Thrift.Parser.FileGroup do
   def resolve(%FileGroup{}=group, %Field{type: %StructRef{}=ref}=field) do
     %Field{field | type: resolve(group, ref)}
   end
+  def resolve(%FileGroup{}=group, %Field{type: {:list, elem_type}}=field) do
+    %Field{field | type: {:list, resolve(group, elem_type)}}
+  end
+  def resolve(%FileGroup{}=group, %Field{type: {:set, elem_type}}=field) do
+    %Field{field | type: {:set, resolve(group, elem_type)}}
+  end
+  def resolve(%FileGroup{}=group, %Field{type: {:map, {key_type, val_type}}}=field) do
+    %Field{field | type: {:map, {resolve(group, key_type), resolve(group, val_type)}}}
+  end
 
   def resolve(%FileGroup{resolutions: resolutions}, %StructRef{referenced_type: type_name}) do
     resolutions[type_name]
   end
-
-  def resolve(%FileGroup{resolutions: resolutions}, path) when is_atom(path) do
-    resolutions[path]
+  for type <- [:bool, :byte, :i8, :i16, :i32, :i64, :double, :string, :binary] do
+    def resolve(_, unquote(type)), do: unquote(type)
   end
 
+  def resolve(%FileGroup{resolutions: resolutions}, path) when is_atom(path) do
+    # this can resolve local mappings like :Weather or
+    # remote mappings like :"common.Weather"
+    resolutions[path]
+  end
   def resolve(_, other) do
     other
   end
-
 
   def dest_module(file_group, %Struct{name: name}) do
     dest_module(file_group, name)
--- a/lib/parser/file_group.ex
+++ b/lib/parser/file_group.ex
@@ -83,6 +83,9 @@ defmodule Thrift.Parser.FileGroup do
                ns_mappings: ns_mappings}
   end
 
+  for type <- [:bool, :byte, :i8, :i16, :i32, :i64, :double, :string, :binary] do
+    def resolve(_, unquote(type)), do: unquote(type)
+  end
   def resolve(%FileGroup{}=group, %Field{type: %StructRef{}=ref}=field) do
     %Field{field | type: resolve(group, ref)}
   end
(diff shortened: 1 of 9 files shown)


test/support/lib/thrift_test_case.ex
```diff
@@ -80,7 +103,122 @@ defmodule ThriftTestCase do
...
+          iolist_struct = with({:ok, tf} <- :thrift_memory_buffer.new_transport_factory(),
+                               {:ok, pf} <- :thrift_binary_protocol.new_protocol_factory(tf, []),
+                               {:ok, binary_protocol} <- pf.()) do
+
+            {proto, :ok} = :thrift_protocol.write(binary_protocol, {struct_info, record})
+            {_, data} = :thrift_protocol.flush_transport(proto)
+            data
+          end
+
+          if Keyword.get(opts, :convert_to_binary, true) do
```
pguillory: Any reason we wouldn't always want this true in tests?

test/protocols/binary_test.exs
```diff
@@ -1,80 +1,183 @@
...
-  import ParserUtils
+  def round_trip_struct(data, serializer_mf, deserializer_mf) do
+    {serializer_mod, serializer_fn} = serializer_mf
+    {deserializer_mod, deserializer_fn} = deserializer_mf
 
-  setup_all do
-    @thrift_file_path
-    |> parse_thrift
-    |> compile_module
+    serialized = :erlang.apply(serializer_mod, serializer_fn, [:struct, data])
```
pguillory: Why does it require that first :struct parameter? Is it an artifact of the implementation? If so, I think we should have serialize/1 as the exposed API people code against and defp the rest. I.e.:
```
def serialize(struct) do
  serialize(:struct, struct)
end
defp serialize(:struct, struct) do
  # ...
end
```

test/support/lib/thrift_test_case.ex
```diff
@@ -80,7 +103,122 @@ defmodule ThriftTestCase do
...
+      quote do
+        Record.defrecord unquote(underscored_record_name), unquote(fields)
+        def unquote(new_fn_name)(opts \\ []) do
+          unquote_splicing(variable_assigns)
+          record = unquote(underscored_record_name)(unquote(kw_match))
+          :erlang.setelement(1, record, unquote(record_name))
+        end
+
+        def unquote(serialize_fn_name)({unquote(underscored_record_name), unquote_splicing(match)}=record, opts \\ []) do
+          record = :erlang.setelement(1 , record, unquote(record_name))
```
pguillory: Whitespace. `1 ,` => `1,`

test/protocols/binary_test.exs
```diff
@@ -1,80 +1,183 @@
...
+    EVIL = 0x20,
+  }
+
+  struct StructWithEnum {
+    1: Status status
+  }
+  """
+  thrift_test "encoding enums" do
+    encoder = {StructWithEnum.BinaryProtocol, :serialize}
+    decoder = {Erlang.Enums, :deserialize_struct_with_enum}
```
pguillory: I wonder if namespacing will be an issue. The Elixir stuff is namespaced to this test module and aliased. I suppose you could do the same with the generated Erlang.* modules but the Erlang code generated by Thrift won't have any such concept. Should we just not worry about it for now?

test/support/lib/thrift_test_case.ex
scohen: I don't know, figured we'd leave it in for completeness.

test/protocols/binary_test.exs
scohen: It's so you can serialize arbitrary things. 
```elixir
serialize(:struct, %MyStruct{})
serialize({:list, :struct}, [%MyStruct])
serialize({:list, i32}, [1, 2, 3, 4])
...etc
```
We can add the serialize/1, but this isn't germane to this PR.

test/protocols/binary_test.exs
scohen: I struggled with namespacing already a little bit (and of tests with user.ex defined stomping over each other. There's some inflexibility about having each file define a single module unless you obey default namespaces. Aside from good taste, there should be nothing preventing us from having `core.User` and `common.User`, but the default generator won't allow this.

lib/parser/file_group.ex
```diff
@@ -86,20 +86,32 @@ defmodule Thrift.Parser.FileGroup do
...
   end
+  def resolve(%FileGroup{}=group, %Field{type: {:list, elem_type}}=field) do
+    %Field{field | type: {:list, resolve(group, elem_type)}}
+  end
+  def resolve(%FileGroup{}=group, %Field{type: {:set, elem_type}}=field) do
+    %Field{field | type: {:set, resolve(group, elem_type)}}
+  end
+  def resolve(%FileGroup{}=group, %Field{type: {:map, {key_type, val_type}}}=field) do
+    %Field{field | type: {:map, {resolve(group, key_type), resolve(group, val_type)}}}
+  end
```
jparise: Remove this space to maintain the function grouping?

lib/protocols/binary.ex
```diff
@@ -43,7 +44,8 @@ defmodule Thrift.Protocols.Binary do
       defmodule BinaryProtocol do
         unquote(primitive_serializers)
         unquote(generate_serializer(file_group, struct))
-        unquote(Deserializer.struct_deserializer(struct, name, file_group))
+        # Commented out due to compilation failures
+        #unquote(Deserializer.struct_deserializer(struct, name, file_group))
```
jparise: Is this something that will be fixed soon?

test/protocols/binary_test.exs
```diff
@@ -1,80 +1,183 @@
...
+
+    assert {:StructWithEnum, 1} == round_trip_struct(StructWithEnum.new, encoder, decoder)
```
jparise: Might be nice to define an use an `assert_roundtrip` helper function.

scohen: hmm, `assert_roundtrip` would have to be a macro if I'm reading you correctly. What were you thinking?

jparise: Yes, but only if it actually improve readability.

scohen:
@jparise I'm not worried about the coverage; we're not running all tests now due to the deserializer tests and the full thrift integ tests not running.

PR #46, opened 2016-11-21
Sort defs to prevent compile warning

pguillory:
Sometimes deserialize* methods are generated out of the order expected by Elixir. This change flattens nested blocks and then sorts them to eliminate the error.

### DIFF
--- a/lib/generator/utils.ex
+++ b/lib/generator/utils.ex
@@ -1,16 +1,49 @@
 defmodule Thrift.Generator.Utils do
+  @moduledoc """
+  Collection of utilities for working with generated code.
+  """
+
+  @doc """
+  When nesting a quote with multiple defs into another quote, the defs end up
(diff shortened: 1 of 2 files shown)


jparise:
Could you add documentation for your new functions? I think that will make their purpose a lot clearer (beyond the scope of this commit description).

lib/generator/utils.ex
```diff
@@ -0,0 +1,28 @@
+defmodule Thrift.Generator.Utils do
+  def merge_blocks([{:__block__, _, contents} | rest]) do
+    merge_blocks(contents) ++ merge_blocks(rest)
+  end
+
```
jparise: I've been following this style suggestion re: grouping function definitions:

https://github.com/rrrene/elixir-style-guide#group-function-definitions

PR #50, opened 2016-11-22
Improve generated notes

pguillory:
Includes field ID in the notes now. Also prints the type better.

### DIFF
--- a/lib/generator/models.ex
+++ b/lib/generator/models.ex
@@ -122,7 +122,7 @@ defmodule Thrift.Generator.Models do
         _ = unquote "Auto-generated Thrift #{label} #{struct.name}"
         unquote_splicing(for field <- struct.fields do
           quote do
-            _ = unquote "#{field.name} #{inspect field.type}"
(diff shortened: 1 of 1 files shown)


pguillory:
Merging it in. This is straightforward.

lib/generator/models.ex
```diff
@@ -170,4 +170,25 @@ defmodule Thrift.Generator.Models do
...
+    "map<#{to_thrift key_type, file_group},#{to_thrift val_type, file_group}>"
+  end
+  def to_thrift({:set, element_type}, file_group) do
+    "set<#{to_thrift element_type, file_group}>"
+  end
+  def to_thrift({:list, element_type}, file_group) do
+    "list<#{to_thrift element_type, file_group}>"
+  end
+  def to_thrift(%Thrift.Parser.Models.TEnum{name: name}, _file_group) do
+    "#{name}"
```
jparise: Can't this just be `name` (and below, as well)? That's preferable to building a new string object as a result of the interpolation.

PR #51, opened 2016-11-22
Benchmark serialize2

pguillory:
This commit adds the alternate serialize implementation to the performance benchmark.

### DIFF
--- a/bench/binary_protocol.exs
+++ b/bench/binary_protocol.exs
@@ -49,16 +49,30 @@ defmodule BinaryProtocolBenchmark do
   end
 
 
-  bench "elixir serializtion (converted to binary)" do
+  bench "elixir serialization (converted to binary)" do
     for user <- bench_context[:elixir_users] do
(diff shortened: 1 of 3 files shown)


pguillory:
Got these results on my laptop.
```
## BinaryProtocolBenchmark
benchmark name                               iterations   average time 
elixir serialization2 (left as IOList)            10000   320.65 µs/op
elixir serialization (left as IOList)              5000   345.50 µs/op
elixir serialization2 (converted to binary)        5000   463.54 µs/op
elixir serialization (converted to binary)         5000   617.99 µs/op
erlang serialization left as IOList                 500   4761.11 µs/op
erlang serialization (converted to binary)          500   5198.23 µs/op
```

scohen:
Hmm... I don't think it's any faster; I was seeing those speeds for the old implementation. It does look like it's faster to convert to a binary, but that shouldn't matter.

PR #52, opened 2016-11-23
Improve deserialization test coverage

pguillory:
Focused on maps/sets/lists of maps/sets/lists.


scohen:
I can't say I'm a fan of this testing approach; While it's easy to write, it's impossible to tell if your assertions match reality or not.

pguillory:
My view is that it's testing at an appropriate interface. If you're implementing a binary protocol, a unit test is going to talk about bytes. These tests say it meets the spec. We also want to know if it's compatible with other Thrift implementations that claim to meet the spec, such as the Erlang one. That's an integration test though. We want both types of tests.

Issue #56, opened 2016-11-30
Thrift generator should be able to generate a single module


scohen:
Right now it generates an entire directory. Instead, we should be able to point it at a single thrift file and it should figure out how to generate that one file and its dependencies.

```elixir
mix thrift.generate thrift/my_service.thrift
```

PR #59, opened 2016-11-30
Support file argument to thrift.generate

pguillory:
This gives the user more fine grained control over which .thrift files are
processed. For instance, a directory might contain many files but you only want
one or a few.

### DIFF
--- a/lib/mix/tasks/thrift.generate.ex
+++ b/lib/mix/tasks/thrift.generate.ex
@@ -4,22 +4,39 @@ defmodule Mix.Tasks.Thrift.Generate do
   @moduledoc """
   Generate Elixir modules from Thrift schema definitions.
 
-  ## Command line options
+  Syntax:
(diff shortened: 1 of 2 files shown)


pguillory:
Fixes #56.

lib/mix/tasks/thrift.generate.ex
```diff
@@ -4,22 +4,39 @@ defmodule Mix.Tasks.Thrift.Generate do
...
   def run(args) do
-    {opts, _, _} = OptionParser.parse(args)
-
-    thrift_dir = Keyword.get(opts, :thrift_dir, "thrift")
+    {opts, args, _} = OptionParser.parse(args)
+    if args == [] do
+      print_help
+      exit :normal
+    end
+    [input | _] = args
```
jparise: Or `input = hd(args)` because you've already covered the empty list case.

PR #62, opened 2016-12-01
Added support for ordinals

scohen:
It's useful to be able to convert a value to an ordinal, and to
enumerate the ordinals in an enum. This PR adds both.

@jparise @pguillory

### DIFF
--- a/lib/thrift/generator/enum_generator.ex
+++ b/lib/thrift/generator/enum_generator.ex
@@ -31,7 +31,7 @@ defmodule Thrift.Generator.EnumGenerator do
         unquote_splicing(macro_defs)
 
         unquote_splicing(ordinal_defs)
-        def ordinal(_), do: :"!!INVALID"
+        def ordinal(_), do: nil
 
         def ordinals, do: unquote(ordinals)
 
--- a/lib/thrift/generator/enum_generator.ex
+++ b/lib/thrift/generator/enum_generator.ex
@@ -2,7 +2,7 @@ defmodule Thrift.Generator.EnumGenerator do
 
   def generate(name, enum) do
     macro_defs = Enum.map(enum.values, fn {key, value} ->
-      macro_name = to_ordinal(key)
+      macro_name = to_name(key)
       quote do
         defmacro unquote(Macro.var(macro_name, nil)), do: unquote(value)
       end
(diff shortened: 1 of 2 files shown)


lib/thrift/generator/enum_generator.ex
```diff
@@ -1,23 +1,47 @@
...
+    |> Keyword.keys
+    |> Enum.map(&to_ordinal/1)
+
     quote do
       defmodule unquote(name) do
         @moduledoc unquote("Auto-generated Thrift enum #{enum.name}")
         unquote_splicing(macro_defs)
+
+        unquote_splicing(ordinal_defs)
+        def ordinal(_), do: :"!!INVALID"
```
jparise: `:invalid_ordinal` would be more idiomatic, but if you're trying to avoid conflicting with a possible atom value, maybe this should just be `nil`.

There's also an argument for explicitly `raise`ing or not defining this function at all and letting the caller run into a `MatchError`.

scohen: i struggled with that, none of the options are good. I think nil might be the best.
 * There won't be a match error if the function isn't defined, it'll be a NoFunctionClauseError and it could be confusing.
 * raise -ing will almost certainly cause runtime errors that will be much more severe than they should be (oh, someone sent you a new value that you can't process for a single event? die.)
I'll switch to nil.

scohen: Oh right, nil will prevent you from defining an enum like:
```thrift
enum ResponseType {
   NIL 
   OBJECT
   STRING
}
```

scohen:
also, @dantswain, if you have ideas.

lib/thrift/generator/enum_generator.ex
jparise: Maybe something like this then?

```elixir
def ordinal(value), do: String.to_atom("undefined_enum_#{value}")
```

lib/thrift/generator/enum_generator.ex
```diff
@@ -1,23 +1,47 @@
...
+        def ordinals, do: unquote(ordinals)
+
         unquote_splicing(member_defs)
         def member?(_), do: false
       end
     end
   end
+
+  defp to_ordinal(key) do
+    key |> to_string |> String.downcase |> String.to_atom
```
dantswain: Did you decide against converting camel to snake case?  If you wanted to do that, `Macro.camelize/1` will do most of the work.

lib/thrift/generator/enum_generator.ex
```diff
@@ -1,23 +1,47 @@
...
+
     quote do
       defmodule unquote(name) do
         @moduledoc unquote("Auto-generated Thrift enum #{enum.name}")
         unquote_splicing(macro_defs)
+
+        unquote_splicing(ordinal_defs)
+        def ordinal(_), do: nil
+
+        def ordinals, do: unquote(ordinals)
```
pguillory: This will conflict with an enum of the same name.

lib/thrift/generator/enum_generator.ex
pguillory: It's not converting between CamelCase and snake_case, it's converting between UPPER_SNAKE_CASE and lower_snake_case.

pguillory:
+1 to not defining `ordinal(_)` and letting it fail to match.

scohen:
The consequences of having an outdated thrift schema will be crashes if we raise an exception or fail to match. That seems excessive.

scohen:
Also, we use this all over the place, I need this feature.

lib/thrift/generator/enum_generator.ex
```diff
@@ -1,23 +1,56 @@
...
+
     quote do
       defmodule unquote(name) do
         @moduledoc unquote("Auto-generated Thrift enum #{enum.name}")
         unquote_splicing(macro_defs)
+
+        unquote_splicing(value_to_name_defs)
+        def value_to_name(_), do: {:error, :invalid_enum_value}
+
+        unquote_splicing(bang_value_to_name_defs)
```
jparise: Can you now define a single generic `value_to_name!` like this?

```elixir
def value_to_name!(value) do
  {:ok, name} = value_to_name(value)
  name
end
```

That would simplify this code quite a bit.

scohen: If you think a match error is better than a FunctionClauseError.

dantswain:
:+1: would it be worth adding `name_to_value`?

PR #63, opened 2016-12-05
Return :error on deserialization failure

pguillory:


lib/thrift/generator/struct_binary_protocol.ex
```diff
@@ -154,17 +155,25 @@ defmodule Thrift.Generator.StructBinaryProtocol do
   def field_deserializer(:string, field, name, _file_group) do
     quote do
       defp unquote(name)(<<11, unquote(field.id)::16-signed, string_size::32-signed, rest::binary>>, acc) do
-        <<value::binary-size(string_size), rest::binary>> = rest
-        unquote(name)(rest, %{acc | unquote(field.name) => value})
+        case rest do
+          <<value::binary-size(string_size), rest::binary>> ->
+            unquote(name)(rest, %{acc | unquote(field.name) => value})
+          _ ->
+            :error
```
scohen: Why not `{:error, :deserialization_failed}` ?

pguillory: You'd get a match error and a stack trace. If the stack trace leads you to code like:
```
{struct, ""} = MyStruct.deserialize(binary)
```
Whether that fails to match `:error` or `{:error, :deserialization_failed}`, it's the same amount of information either way. Is there more information we could provide? For instance, `{:error, :missing_required_field}`, or `{:error, :invalid_field_type}`.

If we return a `{:error, reason}` form of response, we should switch the successful response to `{:ok, struct, rest_of_binary}` to avoid someone matching `{struct, rest_of_binary}` and inadvertently getting error responses.


If there are multiple types of errors and users need to disambiguate them, I could see returning a `{:error, reason}` type of response, and then probably also a `{:ok, struct, rest}` form of struct on success. `:deserialization_failed` doesn't really help us debug *why* it failed.

scohen:
Imagine I'm writing a server on top of this. I think having `{:error, :decode_failed}` is better, since the server could have `{:error, :tcp_closed}`.
`:error` doesn't give me a lot to go on.

pguillory:
Spoke with scohen. We're going to change the deserialize response to `{:ok, struct, rest_of_binary} | {:error, reason}`. I'll go ahead and merge this and make that change in another commit.

PR #66, opened 2016-12-09
Skip unknown fields

### LINKED ISSUE #64: Unsent optional fields crash the deserializer
Say there are two versions of a thrift spec. One has a field and one doesn't. If we generate models on the spec without the field, it won't deserialize a message that contains the field. 

The correct behavior is skipping that field.

scohen:
If you have an old version of a schema, and are sent a message with
fields from a newer version, you should ignore them and continue
processing.

The old behavior was to crash, which causes problems in production
systems

This fixes #64 

@jparise @pguillory

### DIFF
--- a/lib/thrift/generator/struct_binary_protocol.ex
+++ b/lib/thrift/generator/struct_binary_protocol.ex
@@ -130,9 +130,8 @@ defmodule Thrift.Generator.StructBinaryProtocol do
       defp skip_field(<<_::size(16), rest::binary>>, unquote(@i16)), do: rest
       defp skip_field(<<_::size(32), rest::binary>>, unquote(@i32)), do: rest
       defp skip_field(<<_::size(64), rest::binary>>, unquote(@i64)), do: rest
-      defp skip_field(<<string_size::32-signed, rest::binary>>, unquote(@string)) do
-        << _::binary-size(string_size), remainder::binary>> = rest
-        remainder
+      defp skip_field(<<string_size::32-signed, _::binary-size(string_size), rest::binary>>, unquote(@string)) do
+        rest
       end
 
       # sets and lists
@@ -238,8 +237,12 @@ defmodule Thrift.Generator.StructBinaryProtocol do
   end
   def field_deserializer(:string, field, name, _file_group) do
     quote do
-      defp unquote(name)(<<unquote(@string), unquote(field.id)::16-signed, string_size::32-signed, rest::binary>>, acc) do
-        <<value::binary-size(string_size), rest::binary>> = rest
+      defp unquote(name)(<<unquote(@string),
+                         unquote(field.id)::16-signed,
+                         string_size::32-signed,
+                         value::binary-size(string_size),
+                         rest::binary>>, acc) do
+
         unquote(name)(rest, %{acc | unquote(field.name) => value})
       end
     end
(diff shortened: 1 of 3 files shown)


lib/thrift/generator/struct_binary_protocol.ex
```diff
@@ -93,16 +106,88 @@ defmodule Thrift.Generator.StructBinaryProtocol do
...
+      defp skip_map_field(<<rest::binary>>, key_type, val_type, remaining) do
+        rest
+        |> skip_field(key_type)
+        |> skip_field(val_type)
+        |> skip_map_field(key_type, val_type, remaining - 1)
+      end
+
+      defp skip_field(<<rest::binary>>, unquote(@struct)) do
+        skip_struct(rest)
+      end
```
pguillory: All the `skip_field` clauses should be grouped together. Preferably in order of field type ordinal (i.e. from bool=2 to list=15), as that is the pattern used elsewhere.

lib/thrift/generator/struct_binary_protocol.ex
```diff
@@ -93,16 +106,88 @@ defmodule Thrift.Generator.StructBinaryProtocol do
...
+      defp skip_field(<<string_size::32-signed, rest::binary>>, unquote(@string)) do
+        << _::binary-size(string_size), remainder::binary>> = rest
+        remainder
+      end
+
+      # sets and lists
+      defp skip_field(<<elem_type, length::size(32), rest::binary>>, field_type)
+        when field_type == unquote(@set) or field_type == unquote(@list) do
+        skip_list_field(rest, elem_type, length)
+      end
```
pguillory: Recommend just making this two clauses, for consistency.
```
defp skip_field(<<...>>, unquote(@set)), do: ...
defp skip_field(<<...>>, unquote(@list)), do: ...
```

lib/thrift/generator/struct_binary_protocol.ex
```diff
@@ -93,16 +106,88 @@ defmodule Thrift.Generator.StructBinaryProtocol do
...
+
+      defp skip_list_field(<<rest::binary>>, _, 0) do
+        rest
+      end
+      defp skip_list_field(<<rest::binary>>, elem_type, remaining) do
+        rest
+        |> skip_field(elem_type)
+        |> skip_list_field(elem_type, remaining - 1)
+      end
```
pguillory: Recommend naming it `skip_list_element`. The list isn't necessarily a field, it might be an element in another list, etc. And the next one maybe `skip_map_entry`? And for `skip_field`, how about `skip_value`, since it skips any type of value?

lib/thrift/generator/struct_binary_protocol.ex
```diff
@@ -93,16 +106,88 @@ defmodule Thrift.Generator.StructBinaryProtocol do
...
+        skip_struct(rest)
+      end
+
+      defp skip_struct(<<type, _id::16-signed, rest::binary>>) do
+        case skip_field(rest, type) do
+          <<0::size(8), rest::binary>> ->
+            # we have a struct stop
+            rest
+
+          remaining_fields when is_binary(remaining_fields) ->
```
pguillory: How about `<<rest::binary>>`?

scohen: I didn't want to duplicate what was above.

lib/thrift/generator/struct_binary_protocol.ex
scohen: Alright, but it seems verbose.

PR #79, opened 2016-12-24
Replace some Enum.map calls with for loops

jparise:
We weren't using the `Enum.map/2` results here so `for` loops are more
appropriate.


pinterest-ebert:

* 1 fixed issue! 🎉

You can see more details about this review at https://ebertapp.io/github/pinterest/elixir-thrift/pulls/79.

PR #81, opened 2016-12-25
Added Union support

scohen:
Unions are basically like structs, but throw an exception if more than
one field is set.

This adds support for them, I want some feedback on the implementation, which piggybacks heavily on the struct implementation.

### DIFF
--- a/lib/thrift/exceptions.ex
+++ b/lib/thrift/exceptions.ex
@@ -21,12 +21,4 @@ defmodule Thrift.Union.TooManyFieldsSetException do
   field is set.
   """
   defexception message: nil, set_fields: nil
-
-  def new(set_fields) do
-    field_names = Enum.map(set_fields, &elem(&1, 0))
-    %__MODULE__{
-      message: "Thrift union has more than one field set",
-      set_fields: field_names
-    }
-  end
 end
--- a/lib/thrift/exceptions.ex
+++ b/lib/thrift/exceptions.ex
@@ -14,3 +14,7 @@ defmodule Thrift.TApplicationException do
   def exception_type(10), do: :unsupported_client_type
   def exception_type(_), do: :unknown
 end
+
+defmodule Thrift.UnionFieldsSetException do
+  defexception message: nil, set_fields: nil
+end
--- a/lib/thrift/exceptions.ex
+++ b/lib/thrift/exceptions.ex
@@ -15,6 +15,14 @@ defmodule Thrift.TApplicationException do
   def exception_type(_), do: :unknown
 end
 
-defmodule Thrift.UnionFieldsSetException do
+defmodule Thrift.Union.TooManyFieldsSetException do
   defexception message: nil, set_fields: nil
+
+  def new(set_fields) do
+    field_names = Enum.map(set_fields, &elem(&1, 0))
+    %__MODULE__{
+      message: "Thrift union has more than one field set",
+      set_fields: field_names
+    }
+  end
 end
--- a/lib/thrift/exceptions.ex
+++ b/lib/thrift/exceptions.ex
@@ -16,6 +16,10 @@ defmodule Thrift.TApplicationException do
 end
 
 defmodule Thrift.Union.TooManyFieldsSetException do
+  @moduledoc """
+  This exception occurs when a Union is serialized and more than one
+  field is set.
+  """
   defexception message: nil, set_fields: nil
 
   def new(set_fields) do
--- a/lib/thrift/generator/struct_binary_protocol.ex
+++ b/lib/thrift/generator/struct_binary_protocol.ex
@@ -46,6 +46,7 @@ defmodule Thrift.Generator.StructBinaryProtocol do
     Struct,
     StructRef,
     TEnum,
+    Union,
   }
 
   @bool 2
@@ -56,26 +57,49 @@ defmodule Thrift.Generator.StructBinaryProtocol do
   @i64 10
   @string 11
   @struct 12
+  @union 12
   @map 13
   @set 14
   @list 15
 
   @doc """
-  Generate a deserializer for a Thrift struct or exception.
+  Generate a deserializer for a Thrift struct, union or exception.
 
   At the moment it also generates an experimental serializer that may be faster.
   """
-  def struct_deserializer(%{fields: fields}, name, file_group) do
+  def struct_deserializer(%{fields: fields} = model, name, file_group) do
     fields = Enum.reject(fields, &(&1.type == :void))
+
+    struct_check = if model.__struct__ == Thrift.Parser.Models.Union do
+      quote do
+        set_fields = to_serialize
+        |> Map.delete(:__struct__)
+        |> Enum.reject(fn {_, val} -> is_nil(val) end)
+        case set_fields do
+          [] ->
+            :ok
+
+          [_] ->
+            :ok
+
+          set_fields ->
+            raise Thrift.Union.TooManyFieldsSetException.new(set_fields)
+        end
+      end
+    else
+      quote do
+        _ = to_serialize
+      end
+    end
 
     struct_matcher = case fields do
                        [] ->
-                         quote do: %unquote(name){}
+                         quote do: %unquote(name){} = to_serialize
                        f when is_list(f) ->
                          field_matchers = Enum.map(f, fn %Field{name: name} ->
                            {name, Macro.var(name, nil)}
                          end)
-                         quote do: %unquote(name){unquote_splicing(field_matchers)}
+                         quote do: %unquote(name){unquote_splicing(field_matchers)} = to_serialize
                      end
 
     field_serializers = fields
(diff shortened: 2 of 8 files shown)


lib/thrift/generator/struct_binary_protocol.ex
```diff
@@ -100,8 +124,9 @@ defmodule Thrift.Generator.StructBinaryProtocol do
     |> Utils.merge_blocks
 
     quote do
-      def serialize(unquote(struct_matcher)) do
-        unquote([field_serializers, <<0>>] |> Utils.merge_binaries)
+     def serialize(unquote(struct_matcher)) do
```
jparise: It looks like these next three lines are mis-indented by one character.

lib/thrift/exceptions.ex
```diff
@@ -14,3 +14,15 @@ defmodule Thrift.TApplicationException do
   def exception_type(10), do: :unsupported_client_type
   def exception_type(_), do: :unknown
 end
+
+defmodule Thrift.Union.TooManyFieldsSetException do
```
jparise: Add a `@moduledoc` that describes this exception?

lib/thrift/protocols/binary.ex
```diff
@@ -1,4 +1,8 @@
 defmodule Thrift.Protocols.Binary do
+  alias Thrift.Parser.Models.Union
+  alias Thrift.Union.TooManyFieldsSetException
```
pguillory: Unused aliases?

lib/thrift/parser/models.ex
```diff
@@ -254,6 +254,31 @@ defmodule Thrift.Parser.Models do
...
+            :ok
+
+          [_] ->
+            :ok
+
+          set_fields ->
+            raise Thrift.Union.TooManyFieldsSetException.new(set_fields)
+        end
+      end
+    end
```
pguillory: This is code generation stuff and not related to parsing. It would be more at home somewhere under Thrift.Generator than Thrift.Parser. Maybe we should generate a `valid?` method on union models that does this check.

I'm concerned about the performance impact of doing a Map.delete and Enum.reject every time we serialize a union. I suspect it would be significantly faster to use the match on the map itself to enforce that only one field is set. We're already doing that match anyway.

```elixir
union MyUnion {1: string foo, 2: string bar, 3: string baz}

def serialize(%MyUnion{foo: nil, bar: nil, baz: nil})
def serialize(%MyUnion{foo: foo, bar: nil, baz: nil})
def serialize(%MyUnion{foo: nil, bar: bar, baz: nil})
def serialize(%MyUnion{foo: nil, bar: nil, baz: baz})
```

If you submit this PR with the current approach, would you mind if I explored such a refactoring in a subsequent PR?

lib/thrift/exceptions.ex
```diff
@@ -14,3 +14,19 @@ defmodule Thrift.TApplicationException do
...
+  """
+  defexception message: nil, set_fields: nil
+
+  def new(set_fields) do
+    field_names = Enum.map(set_fields, &elem(&1, 0))
+    %__MODULE__{
+      message: "Thrift union has more than one field set",
+      set_fields: field_names
+    }
+  end
```
pguillory: This method is only used in one place. It also has high coupling with that location in the way it takes set_fields but doesn't use all the tuple elements. Recommend inlining it.

test/generator/binary_protocol_test.exs
```diff
@@ -221,6 +222,53 @@ defmodule Thrift.Generator.BinaryProtocolTest do
...
+    3: string string_field,
+    4: list<i16> list_field;
+  }
+
+  struct UStruct {
+    1: Union my_union,
+    2: list<Union> u_list,
+    3: set<Union> u_set,
+    4: map<Union, Union> u_map,
+  }
```
pguillory: These fields should be re-ordered as [value, map, set, list] to maintain consistency with the rest of these tests.

test/generator/binary_protocol_test.exs
```diff
@@ -221,6 +222,53 @@ defmodule Thrift.Generator.BinaryProtocolTest do
...
+
+    assert_raise TooManyFieldsSetException, fn ->
+      Binary.serialize(:union, %Union{int_field: 205, list_field: [1, 2]})
+    end
+  end
+
+  thrift_test "structs can have unions" do
+    assert_serializes %UStruct{},                                     <<0>>
+    assert_serializes %UStruct{my_union: %Union{int_field: 2}},       <<12, 0, 1, 10, 0, 1, 0, 0, 0, 0, 0, 0, 0, 2, 0, 0>>
+
```
pguillory: Extra newline.

test/generator/binary_protocol_test.exs
```diff
@@ -221,6 +222,53 @@ defmodule Thrift.Generator.BinaryProtocolTest do
...
+    assert_serializes %UStruct{u_list: [%Union{int_field: 1291}]},    <<15, 0, 2, 12, 0, 0, 0, 1, 10, 0, 1, 0, 0, 0, 0, 0, 0, 5, 11, 0, 0>>
+    assert_serializes %UStruct{u_set: MapSet.new([%Union{int_field: 2239}])},
+                                                                      <<14, 0, 3, 12, 0, 0, 0, 1, 10, 0, 1, 0, 0, 0, 0, 0, 0, 8, 191, 0, 0>>
+    assert_serializes %UStruct{u_map: %{}},                           <<13, 0, 4, 12, 12, 0, 0, 0, 0, 0>>
+    assert_serializes %UStruct{
+      u_map: %{%Union{int_field: 23} => %Union{int_field: 33}}},      <<13, 0, 4, 12, 12, 0, 0, 0, 1, 10, 0, 1, 0, 0, 0, 0, 0, 0, 0, 23, 0, 10, 0, 1, 0, 0, 0, 0, 0, 0, 0, 33, 0, 0>>
```
pguillory: This line is wrapped inconsistently with the other `assert_serializes` invocations in this file.

lib/thrift/parser/models.ex
jparise: I think that's a neat idea.

Building on it, I wonder if there's a maximum union "size" (field count) where it becomes less efficient to pattern match and more efficient to handle within a single function body. Because we control code generation, we could select the more optimal strategy.

dantswain:
Does it make sense to provide a `Thrift.Union` module with some helper functions?

scohen:
@dantswain I think it might, but I don't see how we could enforce its usage. I'm open to ideas.

lib/thrift/parser/models.ex
scohen: >  I suspect it would be significantly faster to use the match on the map itself to enforce that only one field is set.

I'd like to see benchmarks that prove that out before any refactoring is explored. In my testing, together, the delete and reject were single-digit microsecond operations.

In addition, the above approach yields the dreaded and unhelpful "NoFunctionClauseError".

test/generator/binary_protocol_test.exs
scohen: I'm curious as to why the ordering matters. Are the names not clear?

test/generator/binary_protocol_test.exs
scohen: I, for one, found it easier to read this way, since this is the longest struct definition in the file. 

I think we're focusing a lot on whitespace and not as much as we should on content. Unlike python, Elixir doesn't have a set way to format your code and instead relies on the programmer to make things look pretty. 

Rather than saying "This doesn't look like the rest of the file", you should instead ask if it's readable and does it convey the intent properly.

dantswain:
@scohen I don't think it would need to be enforced, really.  The implementation is still just a struct, and the module is just there to help people work with it if they want.  For example I'm thinking something along the lines of

```
# return the value for the only key that is set
def value(union)
# set key to value and return the modified union, raise if any other keys are already set
def set!(union, key, value)
# return true if one or zero values are set
def valid?(union)
```

It would be optional to use these, but they might help bridge the semantic gap introduced by the fact that they're implemented as structs and Elixir has no native notion of a union type.

Totally an optional thing and shouldn't hold up merging this PR.

scohen:
@dantswain Should these be added to each generated Union module?

```elixir
u = %MyUnion{}
u = MyUnion.set!(u, :first_val, 3)
MyUnion.value(u)
=> 3
```

dantswain:
@scohen I was thinking a plain old module unless the performance hit is massive.  One can always go down to the lower level if they want better performance.

test/generator/binary_protocol_test.exs
pguillory: It's helpful when scanning the code with a question in mind like, did we test all the combinations? Val, map, set, list, val, map, set, list...

I can see how it would come across as OCD, but consistency like that is kind to future maintainers of the code.

scohen: I'm under the impression that this code isn't really _scannable_ in any sense, and would be in favor of helper functions to make validations easier for people just coming into the codebase. 

...but we've had that discussion before.

PR #82, opened 2016-12-26
Use `defp` for internal module functions

jparise:


lib/thrift/generator/struct_binary_protocol.ex
```diff
@@ -131,8 +130,7 @@ defmodule Thrift.Generator.StructBinaryProtocol do
     end
   end
 
-  def field_skippers do
-
+  defp field_skippers do
```
pinterest-ebert: Function is too complex (CC is 24, max is 9).

PR #90, opened 2016-12-27
Match unions with zero or one field set

pguillory:
This uses matching in the `serialize` definition to enforce that only unions
with zero or one field can be serialized.

### DIFF
--- a/lib/thrift/generator/struct_binary_protocol.ex
+++ b/lib/thrift/generator/struct_binary_protocol.ex
@@ -67,31 +67,14 @@ defmodule Thrift.Generator.StructBinaryProtocol do
 
   At the moment it also generates an experimental serializer that may be faster.
   """
-  def struct_deserializer(%{fields: fields} = model, name, file_group) do
+  def struct_deserializer(%{fields: fields}, name, file_group) do
     fields = Enum.reject(fields, &(&1.type == :void))
-    struct_matcher = case fields do
-                       [] ->
-                         quote do: %unquote(name){} = to_serialize
-                       f when is_list(f) ->
-                         field_matchers = Enum.map(f, fn %Field{name: name} ->
-                           {name, Macro.var(name, nil)}
-                         end)
-                         quote do: %unquote(name){unquote_splicing(field_matchers)} = to_serialize
-                     end
-
-    field_serializers = fields
-    |> Enum.map(&field_serializer(&1, file_group))
 
     field_deserializers = fields
     |> Enum.map(&field_deserializer(&1.type, &1, :deserialize, file_group))
     |> Utils.merge_blocks
 
     quote do
-      def serialize(unquote(struct_matcher)) do
-        unquote(Union.validator(model, :to_serialize))
-        unquote([field_serializers, <<0>>] |> Utils.merge_binaries)
-      end
-
       def deserialize(binary) do
         deserialize(binary, %unquote(name){})
       end
@@ -190,21 +173,6 @@ defmodule Thrift.Generator.StructBinaryProtocol do
       end
       defp skip_struct(_) do
         :error
-      end
-    end
-  end
-
-  defp field_serializer(%Field{name: name, type: type, id: id}, file_group) do
-    var = Macro.var(name, nil)
-    quote do
-      case unquote(var) do
-        nil ->
-          <<>>
-        _ ->
-          unquote([
-            quote do <<unquote(type_id(type, file_group)), unquote(id) :: size(16)>> end,
-            value_serializer(type, var, file_group)
-          ] |> Utils.merge_binaries |> Utils.simplify_iolist)
       end
     end
   end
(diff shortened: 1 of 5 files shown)


test/support/lib/parser_utils.ex
```diff
@@ -1,5 +1,9 @@
 defmodule User do
-  defstruct is_evil: false, user_id: 0, number_of_hairs_on_head: 0, amount_of_red: 0, nineties_era_color: 0, mint_gum: 0.0, username: "", friends: [], my_map: %{}, blocked_user_ids: MapSet.new(), optional_integers: []
+  defstruct is_evil: false, user_id: 0, number_of_hairs_on_head: 0, amount_of_red: 0, nineties_era_color: 0, mint_gum: 0.0, username: "", friends: [], my_map: %{}, blocked_user_ids: MapSet.new(), optional_integers: [], u: nil
+end
+
+defmodule TestUnion do
```
pinterest-ebert: Modules should have a @moduledoc tag.

pinterest-ebert:

But beware that this branch is **[1 commit behind the `pinterest:thrift_tng` branch](https://github.com/pinterest/elixir-thrift/compare/pguillory:match_union...pinterest:thrift_tng)**, and a review of an up to date branch would produce more accurate results.

You can see more details about this review at https://ebertapp.io/github/pinterest/elixir-thrift/pulls/90.

pguillory:
I ran the included benchmark before and after the changes, and also with a no-op loop. (There's a bit of overhead just looping 1000 times.) It appears to take about 77% as much time to serialize a union after the change as before.

```
before:
serialize union       10000   1439.41 µs/op

after:
serialize union       10000   1131.56 µs/op

noop:
serialize union      100000   95.33 µs/op
```

pguillory:
Here is some example generated code, to get a sense of what's going on.

Before:
```elixir
def(serialize(%TestUnion{s: s, i: i, m: m} = to_serialize)) do
  (
    set_fields = to_serialize |> Map.delete(:__struct__) |> Enum.reject(fn {_, val} -> is_nil(val) end)
    case(set_fields) do
      [] ->
        :ok
      [_] ->
        :ok
      set_fields ->
        field_names = Enum.map(set_fields, &(elem(&1, 0)))
        raise(%Thrift.Union.TooManyFieldsSetException{message: "Thrift union has more than one field set", set_fields: field_names})
    end
  )
  [case(s) do
    nil ->
      <<>>
    _ ->
      [<<11, 1::size(16), byte_size(s)::size(32)>>, s]
  end, case(i) do
    nil ->
      <<>>
    _ ->
      <<10, 2::size(16), i::64-signed>>
  end, case(m) do
    nil ->
      <<>>
    _ ->
      [<<13, 3::size(16), 11, 11, map_size(m)::size(32)>>, for({k, v} <- m) do
        [<<byte_size(k)::size(32)>>, k, <<byte_size(v)::size(32)>>, v]
      end]
  end, <<0>>]
end
```

After:
```elixir
def(serialize(%TestUnion{s: nil, i: nil, m: nil})) do
  <<0>>
end
def(serialize(%TestUnion{s: s, i: nil, m: nil})) do
  [<<11, 1::size(16), byte_size(s)::size(32)>>, s, <<0>>]
end
def(serialize(%TestUnion{i: i, s: nil, m: nil})) do
  [<<10, 2::size(16), i::64-signed>>, <<0>>]
end
def(serialize(%TestUnion{m: m, s: nil, i: nil})) do
  [<<13, 3::size(16), 11, 11, map_size(m)::size(32)>>, for({k, v} <- m) do
    [<<byte_size(k)::size(32)>>, k, <<byte_size(v)::size(32)>>, v]
  end, <<0>>]
end
def(serialize(%TestUnion{} = value)) do
  set_fields = value |> Map.from_struct() |> Enum.flat_map(fn
    {_, nil} ->
      []
    {key, _} ->
      [key]
  end)
  raise(%Thrift.Union.TooManyFieldsSetException{message: "Thrift union has more than one field set", set_fields: set_fields})
end
```

lib/thrift/generator/struct_binary_protocol.ex
```diff
@@ -67,31 +67,14 @@ defmodule Thrift.Generator.StructBinaryProtocol do
 
   At the moment it also generates an experimental serializer that may be faster.
```
jparise: Is this statement still accurate?

lib/thrift/generator/struct_binary_protocol.ex
```diff
@@ -803,6 +771,90 @@ defmodule Thrift.Generator.StructBinaryProtocol do
     |> list_deserializer(name, file_group)
   end
 
+  def struct_serializer(%Union{fields: fields}, name, file_group) do
```
jparise: Let's `@doc` this if it's intended to be part of the exported API.

lib/thrift/generator/struct_binary_protocol.ex
```diff
@@ -803,6 +771,90 @@ defmodule Thrift.Generator.StructBinaryProtocol do
...
+    fields = Enum.reject(fields, &(&1.type == :void))
+
+    single_field_serializers = Enum.map(fields, fn
+      %Field{name: match_name} = match_field ->
+        nil_matchers = Enum.flat_map(fields, fn
```
jparise: How does this compare performance-wise to:

```elixir
nil_matchers = Enum.filter_map(fields,
  fn(%Field{name: match_name}) -> match_name != name end,
  fn(%Field{name: match_name}) -> {match_name, nil} end)
```

I imagine your approach is faster because it can match directly on the name, but `Enum.filter_map/3` uses fewer intermediate lists.

This is just the codegen bit, though, so readability could trump performance.

lib/thrift/generator/struct_binary_protocol.ex
```diff
@@ -803,6 +771,90 @@ defmodule Thrift.Generator.StructBinaryProtocol do
...
+        field_serializers = [required_field_serializer(match_field, file_group)]
+
+        quote do
+          def serialize(%unquote(name){unquote_splicing(field_matchers)}) do
+            unquote([field_serializers, <<0>>] |> Utils.merge_binaries)
+          end
+        end
+    end)
+
+    all_fields_nil = Enum.map(fields, fn
```
jparise: If you need to build this list regardless, you can also do this step earlier and then quickly filter it to produce `nil_matchers` using just:

```elixir
nil_matchers = Enum.reject(all_fields_nil, fn({nil_name, _}) -> nil_name == name end)
```

test/support/lib/parser_utils.ex
jparise: You can just use `@moduledoc  false` here.

lib/thrift/generator/struct_binary_protocol.ex
pguillory: Is there something like package visibility in Elixir? It's def (vs defp) because it's used by another module, but it's not something that end users should be coding against.

lib/thrift/generator/struct_binary_protocol.ex
pguillory: As you said it runs in the codegen phase. I don't think it's really worth benchmarking. I'll switch it though if you favor the other one for readability.

lib/thrift/generator/struct_binary_protocol.ex
pguillory: Yeah I'll do something like this. It's cleaner.

pinterest-ebert:

* 4 fixed issues! 🎉

But beware that this branch is **[1 commit behind the `pinterest:thrift_tng` branch](https://github.com/pinterest/elixir-thrift/compare/pguillory:match_union...pinterest:thrift_tng)**, and a review of an up to date branch would produce more accurate results.

You can see more details about this review at https://ebertapp.io/github/pinterest/elixir-thrift/pulls/90.

lib/thrift/generator/struct_binary_protocol.ex
jparise: No, functions are either exported (public) or not.

We can revisit documentation later if you don't think it adds anything now.

lib/thrift/generator/struct_binary_protocol.ex
jparise: I'll leave it to your preference (and it looks like you'll rewrite it per below anyway). I was mostly just thinking about other ways to write that expression. 😊

PR #92, opened 2016-12-28
More robust handling of void return types

scohen:
The erlang server replies to void responses with <<12, 0, 0, 0, 0>>
which is an empty struct and the stop byte from the respone
struct. Other implementations just reply with the stop byte from the
response struct. We can now parse both.

I also fixed a bug that would occur if the sequence ids or messages
would mismatch. I added tests for all cases.

### DIFF
--- a/test/clients/binary_framed_test.exs
+++ b/test/clients/binary_framed_test.exs
@@ -1,0 +1,63 @@
+defmodule BinaryFramedTest do
+  use ThriftTestCase
+
+  alias Thrift.Clients.BinaryFramed
+
+  @thrift_file name: "void_return.thrift", contents: """
+  service VoidReturns {
+    void my_call(1: i64 id)
+  }
+  """
+  alias BinaryFramedTest.VoidReturns
+  alias Thrift.TApplicationException, as: TAE
+
+  thrift_test "it should be able to deserialize an invalid message" do
+    msg = <<128, 1, 0, 2>>
+    assert {:error, {:cant_decode_message, ^msg}} = BinaryFramed.deserialize_message_reply(msg, "my_call", 2757, VoidReturns.MyCallResponse)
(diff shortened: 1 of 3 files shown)


test/clients/binary_framed_test.exs
```diff
@@ -0,0 +1,63 @@
...
+
+  alias Thrift.Clients.BinaryFramed
+
+  @thrift_file name: "void_return.thrift", contents: """
+  service VoidReturns {
+    void my_call(1: i64 id)
+  }
+  """
+  alias BinaryFramedTest.VoidReturns
+  alias Thrift.TApplicationException, as: TAE
```
pguillory: I would have thought VoidReturns had already been aliased by ThriftTestCase.

All the aliases should be grouped together up at the top. They don't need to be below the `@thrift_file`.

test/clients/binary_framed_test.exs
```diff
@@ -0,0 +1,63 @@
...
+  end
+
+  thrift_test "it should be able to deserialize a message with the wrong method name and sequence id " do
+    msg = <<128, 1, 0, 2, 0, 0, 0, 8, "bad_call", 0, 0, 10, 197, 0>>
+
+    assert {:error, {:exception, ex}} = BinaryFramed.deserialize_message_reply(msg, "my_call", 1234, VoidReturns.MyCallResponse)
+    assert %TAE{type: :sequence_id_and_rpc_name_mismatched} = ex
+  end
+
+  thrift_test "It should be able to deserialize a void message" do
```
jparise: Above you use lowercase "it".

test/clients/binary_framed_test.exs
scohen: It is aliased properly. removing.

PR #97, opened 2016-12-29
Always name the deserialize function `deserialize`

jparise:
The previous code supported variable function names, but in practice we
always named it `deserialize`. This simplifies the code a bit and makes
the output of these macros a bit more obvious.

### DIFF
--- a/lib/thrift/generator/struct_binary_protocol.ex
+++ b/lib/thrift/generator/struct_binary_protocol.ex
@@ -69,7 +69,7 @@ defmodule Thrift.Generator.StructBinaryProtocol do
     fields = Enum.reject(fields, &(&1.type == :void))
 
     field_deserializers = fields
-    |> Enum.map(&field_deserializer(&1.type, &1, :deserialize, file_group))
+    |> Enum.map(&field_deserializer(&1.type, &1, file_group))
     |> Utils.merge_blocks
 
     quote do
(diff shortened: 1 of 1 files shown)


scohen:
LGTM, but this is more preston's purview.

jparise:
@pguillory, is that something you could explore soon? I understand what you have in mind, but I'd also prefer to keep things simple and less speculative until we're able to take that step.

pguillory:
@jparise Yeah I'll look into it.

dantswain:
+1 for simplification

@pguillory How does this change affect the ability to inline a substruct's deserializer?  Or maybe I don't understand :/

scohen:
@dantswain If we have struct a that contains struct b, the deserialization looks like
```elixir
def deserialize(<<struct_a_field_number, rest::binary, struct_a) do 
  {struct_b, leftover} = StructBMod.deserialize(rest)
  deserialize(leftover, %StructAMod{struct_a | struct_b: struct_b})
end
```
but it would be more efficient to somehow inline that call to StructBMod.deserialize, since when we use the binary in the function, we lose the binary's match context. 

In order to keep the match context, you must not modify the binary, it has to be the first argument and the call must be recursive.

It's a tough triumvirate of rules, that.

dantswain:
@scohen I can dig that but I'm not understanding how it's related to the changes in this PR.  I'm probably just missing something and I want to understand.

pguillory:
If we want to inline the deserialization of a struct into some other module, step 1 will be to alter struct_deserializer to take the method name as an argument. Which is pretty easy, because everything that makes an assumption about the name `deserialize` is right there in that one function.

If this commit goes through, the new step 1 will be to revert this commit. I think there's about a 75% chance we'll want to do that in the next few weeks. In addition to structs inside structs, we might want to interleave deserialization of structs and messages, as used by the client/server.

I worked on it actually but hit a roadblock in running the test suite, because we load the generated files one at a time and they refer to structs in other modules that hadn't been loaded yet.

PR #99, opened 2016-12-29
Behaviour module generation

scohen:
For the server side, I imagine that the user would specify a handler
module and pass it into the service. Then it would be called with the
appropriate arguments.
I thought it might be nice to have a behaviour for these modules that
users could be implement.

Behavoiurs here are fully typed with the correct success types.

### DIFF
--- a/lib/thrift.ex
+++ b/lib/thrift.ex
@@ -9,8 +9,8 @@ defmodule Thrift do
   """
 
   @type i8   :: (-128..127)
-  @type double :: float()
   @type i16 :: (-32_768..32_767)
   @type i32 :: (-2_147_483_648..2_147_483_647)
   @type i64 :: (-9_223_372_036_854_775_808..9_223_372_036_854_775_807)
+  @type double :: float()
 end
--- a/lib/thrift.ex
+++ b/lib/thrift.ex
@@ -7,4 +7,10 @@ defmodule Thrift do
 
     * `Thrift.Parser` - functions for parsing `.thift` schema files
   """
+
+  @type i8   :: (-128..127)
+  @type double :: float()
+  @type i16 :: (-32_768..32_767)
+  @type i32 :: (-2_147_483_648..2_147_483_647)
+  @type i64 :: (-9_223_372_036_854_775_808..9_223_372_036_854_775_807)
 end
--- a/lib/thrift/generator/behaviour.ex
+++ b/lib/thrift/generator/behaviour.ex
@@ -1,0 +1,81 @@
+defmodule Thrift.Generator.Behaviour do
+  alias Thrift.Parser.FileGroup
+  alias Thrift.Parser.Models.{
+    Field,
+    Struct,
+    StructRef,
+  }
+
+  def generate(schema, service) do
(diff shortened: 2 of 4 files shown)


lib/thrift/generator/behaviour.ex
```diff
@@ -0,0 +1,81 @@
+defmodule Thrift.Generator.Behaviour do
```
pinterest-ebert: Modules should have a @moduledoc tag.

pinterest-ebert:

You can see more details about this review at https://ebertapp.io/github/pinterest/elixir-thrift/pulls/99.

lib/thrift.ex
```diff
@@ -7,4 +7,10 @@ defmodule Thrift do
 
     * `Thrift.Parser` - functions for parsing `.thift` schema files
   """
+
+  @type i8   :: (-128..127)
+  @type double :: float()
```
jparise: Probably pedantic, but move `double` to the end of the list so it's not in the middle of the integer types?

test/generator/behaviour_test.exs
```diff
@@ -0,0 +1,33 @@
...
+    void numbers(1: byte b, 2: i16 i, 3: i32 eye32, 4: i64 eye64, 5: double dub),
+    void my_set(1: set<string> my_set),
+    void my_list(1: list<string> my_string),
+    void my_map(1: map<string, string> my_map)
+    map<string, bool> my_map2(1: map<string, map<string, string>> my_map)
+    void struct_param(1: S my_struct)
+  }
+  """
+
+  thrift_test "that behaviours exist" do
```
jparise: "that behaviour callbacks exist" ?

test/generator/behaviour_test.exs
```diff
@@ -0,0 +1,33 @@
...
+    behaviour_specs = Behaviour.behaviour_info(:callbacks)
```
jparise: Is the deprecated [`Elixir.Behaviour`](http://elixir-lang.org/docs/stable/elixir/Behaviour.html) module? Or the generated service module? Maybe rename the service to avoid any confusion?

test/generator/behaviour_test.exs
```diff
@@ -0,0 +1,33 @@
...
+
+    assert {:ping, 1} in behaviour_specs
+    assert {:my_set, 1} in behaviour_specs
+    assert {:my_list, 1} in behaviour_specs
+    assert {:my_map, 1} in behaviour_specs
+    assert {:my_map2, 1} in behaviour_specs
+    assert {:numbers, 5} in behaviour_specs
+    assert {:my_bool, 1} in behaviour_specs
+    assert {:struct_param, 1} in behaviour_specs
```
pguillory: Can these be in the same order as their definitions in BehaviorService?

test/generator/behaviour_test.exs
pguillory: *.HandlerBehavior instead of *.Handler.Behavior?

scohen: I have a preference for `Handler.Behaviour`.

PR #103, opened 2016-12-30
Move skip_field/2 to the Binary protocol module

jparise:
We were previously generating private copies of the "skip" functions in
each generated struct module. These functions are generic and have been
moved to the top-level Binary protocol module instead. The reduces the
size of the generated modules by a nice amount.

We only skip unrecognized fields so this code shouldn't be called very
often. In practice, however, we can't predict how often this case will
arise, so the performance of these function is important. Fortunately,
this change doesn't introduce a measurable performance cost.

Lastly, this change adds Binary module documentation and some initial
typespecs.

### DIFF
--- a/lib/thrift/generator/struct_binary_protocol.ex
+++ b/lib/thrift/generator/struct_binary_protocol.ex
@@ -82,91 +82,15 @@ defmodule Thrift.Generator.StructBinaryProtocol do
 
       unquote_splicing(field_deserializers)
 
+      # Skip over unknown fields. This can happen when we encounter serialized
+      # data from a newer schema than our own.
       defp deserialize(<<field_type, _id::16-signed, rest::binary>>, acc) do
-        # this is for unknown fields. This can happen if you have an
-        # outdated schema
         rest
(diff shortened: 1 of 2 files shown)


pinterest-ebert:

* 2 fixed issues! 🎉

You can see more details about this review at https://ebertapp.io/github/pinterest/elixir-thrift/pulls/103.

pguillory:
Ran a benchmark to confirm calling another module is only slightly slower.
```
foo                        1000000   2.71 µs/op
Other.foo                  1000000   2.93 µs/op
apply(Other, :foo, [])     1000000   2.95 µs/op
__MODULE__.foo             1000000   2.96 µs/op
func.()                     200000   7.97 µs/op
apply(mod, :foo, [])        100000   16.96 µs/op
mod.foo                     100000   18.51 µs/op
```

jparise:
@pguillory That's backed up by section 6.2 of the Efficiency Guide: http://erlang.org/doc/efficiency_guide/functions.html

PR #104, opened 2016-12-30
Use pattern matching to serialize bool values

jparise:
This eliminates the private bool_to_int/1 function and lets us return
constant binary result values directly.

We also strictly match on just the `true` and `false` atoms.

There is a nominal performance improvement, but nothing significant.

### DIFF
--- a/lib/thrift/protocols/binary.ex
+++ b/lib/thrift/protocols/binary.ex
@@ -36,10 +36,6 @@ defmodule Thrift.Protocols.Binary do
   def int_type({:set, _}), do: 14
   def int_type({:list, _}), do: 15
 
-  defp bool_to_int(false), do: 0
-  defp bool_to_int(nil), do: 0
-  defp bool_to_int(_), do: 1
-
   defp to_message_type(:call), do: 1
   defp to_message_type(:reply), do: 2
   defp to_message_type(:exception), do: 3
@@ -54,8 +50,11 @@ defmodule Thrift.Protocols.Binary do
     []
   end
   def serialize(:bool, value) do
-    value = bool_to_int(value)
-    <<value::8-signed>>
+    case value do
+      nil   -> <<0::8-signed>>
+      false -> <<0::8-signed>>
+      _     -> <<1::8-signed>>
+    end
   end
   def serialize(:i8, value) do
     <<value::8-signed>>
--- a/lib/thrift/protocols/binary.ex
+++ b/lib/thrift/protocols/binary.ex
@@ -49,9 +49,8 @@ defmodule Thrift.Protocols.Binary do
   def serialize(_, nil) do
     []
   end
-  def serialize(:bool, nil),   do: <<0::8-signed>>
   def serialize(:bool, false), do: <<0::8-signed>>
-  def serialize(:bool, _),     do: <<1::8-signed>>
+  def serialize(:bool, true),  do: <<1::8-signed>>
   def serialize(:i8, value) do
     <<value::8-signed>>
   end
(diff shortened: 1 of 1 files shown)


jparise:
The other way we could do this is to pattern match `nil`, `false`, and `_` directly in the `serialize(:bool, _)` function. Any preference? I suppose that might be even cleaner.

dantswain:
:+1: I like the pattern matching in the function clause.

jparise:
@dantswain yeah, I realize that would be the better approach as I was writing that comment. 😉

lib/thrift/protocols/binary.ex
```diff
@@ -53,10 +49,9 @@ defmodule Thrift.Protocols.Binary do
   def serialize(_, nil) do
     []
   end
-  def serialize(:bool, value) do
-    value = bool_to_int(value)
-    <<value::8-signed>>
-  end
+  def serialize(:bool, nil),   do: <<0::8-signed>>
```
dantswain: Now that I think about it - should this clause ever get called?  If a value is `nil` then it's not set and should be skipped.

pguillory: If a boolean field in a struct is nil, that field would be omitted. That logic isn't represented in this generic serialization because it doesn't handle structs. It passes them off to presumably-generated struct serializers. (See `serialize(:struct, _)` further down.)

If you serialize a list of bools, we would treat a nil in that list as falsey.

```elixir
serialize({:list, :bool}, [nil, false, true]) |> :erlang.iolist_to_binary
# <<2, 0, 0, 0, 3, 0, 0, 1>>
#   |  \--------/  |  |  \- true
#   |      |       |  \---- false
#   |      |       \------- nil
#   |      \--------------- length = 3
#   \---------------------- element type = bool
```

Note that this doesn't currently work because of the `serialize(_, nil)` clause further up.

lib/thrift/protocols/binary.ex
```diff
@@ -53,10 +49,9 @@ defmodule Thrift.Protocols.Binary do
   def serialize(_, nil) do
     []
   end
```
pguillory: This clause should not exist. The way we serialize nil depends on its context.

lib/thrift/protocols/binary.ex
jparise: I think implicit here is whether we want to watch to support "truthy" / "falsey" inputs or if we should explicitly require `true` or `false` atoms (where all other inputs result in an `:error`).

lib/thrift/protocols/binary.ex
jparise: That sounds right to me. I'll remove it in its own PR so we can track the specific motivation.

lib/thrift/protocols/binary.ex
dantswain: @pguillory's comment makes sense to me.  @jparise raises a good point, too.

pguillory: Yeah good point. If we're not sure, it might be better to start strict, true/false only. We can become less strict later without breaking applications.

dantswain: Good point @pguillory.  

Related - some thrift clients (I know the Ruby client does this) provide validation functionality - e.g., are required fields set, is a string field actually a string, etc.  It's slow but it's usually optional.  I typically enable it for testing but disable it in production.  This might be something to add a TODO for later?

lib/thrift/protocols/binary.ex
jparise: This clause is removed in #105.

PR #113, opened 2016-12-30
More docs and typespecs for the Binary module

jparise:


lib/thrift/protocol/binary.ex
```diff
@@ -86,7 +90,13 @@ defmodule Thrift.Protocol.Binary do
     byte_size(name)::32-signed, name::binary, sequence_id::32-signed>>
   end
 
-  def deserialize(:message_begin,<<1::size(1), 1::size(15), _::size(8),
+  @doc """
+  Deserializes a Thrift-encoded binary.
+  """
+  @spec deserialize(atom, binary) ::
+    {:ok, {atom, integer, binary, binary}} |
```
dantswain: Might be helpful to give these names?  At first glance it's not obvious to me what the values are.

jparise:
@dantswain I think it's useful to document this top-level Binary module because it could be useful for people working directly with serialized data. I'm looking though our internal modules now, and it looks like we should `@doc false` a bunch of things to simplify our generated documentation (and public API) however, particularly in the Parser namespace.

dantswain:
@jparse OK. I think it'll be fine as long as the higher-level documentation (e.g., the README) does a good job guiding people to the intended public functions.

jparise:
@dantswain Right. We're basically shipping an IDL parser, protocol-aware functions, and a client/server framework in one package, so we need to make it obvious how all of those pieces should be used.

lib/thrift/protocol.ex
```diff
@@ -0,0 +1,13 @@
+defmodule Thrift.Protocol do
+  @moduledoc """
+  This module contains high-level Thrift protocol types.
+  """
+
+  @typedoc "Thrift data types"
+  @type data_type ::
```
dantswain: Are these specific to protocols or should they be in `Thrift`?

lib/thrift/protocol/binary.ex
```diff
@@ -36,16 +38,25 @@ defmodule Thrift.Protocol.Binary do
...
-  defp to_message_type(:reply), do: 2
-  defp to_message_type(:exception), do: 3
-  defp to_message_type(:oneway), do: 4
+  @spec from_message_type(Protocol.message_type) :: 1..4
+  defp from_message_type(:call),      do: 1
+  defp from_message_type(:reply),     do: 2
+  defp from_message_type(:exception), do: 3
+  defp from_message_type(:oneway),    do: 4
+
+  @spec to_message_type(1..4) :: Protocol.message_type
```
dantswain: Is it worth doing a `@type message_type_id 1 | 2 | 3 | 4`?

lib/thrift/protocol/binary.ex
```diff
@@ -36,16 +38,25 @@ defmodule Thrift.Protocol.Binary do
...
+  defp to_message_type(2), do: :reply
+  defp to_message_type(3), do: :exception
+  defp to_message_type(4), do: :oneway
 
-  defp from_message_type(1), do: :call
-  defp from_message_type(2), do: :reply
-  defp from_message_type(3), do: :exception
-  defp from_message_type(4), do: :oneway
+  @typedoc "Binary protocol message sequence identifier"
+  @type sequence_id :: non_neg_integer
```
dantswain: Does this belong somewhere else?

lib/thrift/protocol.ex
jparise: Yeah, @scohen caught that too. Moving them to `Thrift`.

lib/thrift/protocol/binary.ex
jparise: @scohen, is this definition of sequence ID binary-specific?

PR #114, opened 2016-12-30
Restructure env-specific builds paths

jparise:
Define :test-specific (instead of :prod-specific) lists of build paths.
This fixes an issue where :dev builds would include test code and things
like `mix docs` would parse and document tests.

Because our benchmarks also use test code (e.g. ParserUtils), that
command is runs in the :test environment.


scohen:
Yes, this makes sense.

PR #116, opened 2016-12-30
Binary Framed server

scohen:
This is the implementation of the binary thrift framed server. It features
a configurable acceptor pool and generates code to make serialization and
deserialization automatic.

@fishcakez, if you would be so kind. 

I'm concerned if using a supervisor of acceptor processes is good enough for a pool.


test/servers/binary_framed_test.exs
```diff
@@ -0,0 +1,156 @@
...
+      bool checked_exception() throws (1: TestException ex);
+      bool multiple_exceptions(1: i32 exc_type) throws
+         (1: TestException e, 2: UserNotFound unf, 3: OtherException other);
+      bool server_exception();
+      IdAndName echo_struct(1: IdAndName id_and_name);
+      i64 myCamelCasedFunction(1: string myUserName);
+    }
+  """
+
+  def define_handler do
```
pinterest-ebert: Function is too complex (CC is 13, max is 9).

test/servers/binary_framed_test.exs
```diff
@@ -0,0 +1,156 @@
...
+
+    {:ok, client: client, port: server_port}
+  end
+
+  thrift_test "it can return a simple boolean value", ctx do
+    assert {:ok, true} == Client.Framed.ping(ctx.client)
+  end
+
+  thrift_test "it can throw checked exceptions", ctx do
+    expected_exception = %TestException{message: "Oh noes!", code: 400 }
```
pinterest-ebert: There is no whitespace around parentheses/brackets most of the time, but here there is.

lib/thrift/servers/binary_framed/server.ex
```diff
@@ -0,0 +1,34 @@
...
+  Users should not use this module directly, and instead use one of the generated server modules.
+  """
+
+  @type server_opts :: [
+    worker_count: pos_integer
+  ]
+
+  alias Thrift.Servers.BinaryFramed.AcceptorSupervisor
+
+  defmodule State do
```
pinterest-ebert: Modules should have a @moduledoc tag.

lib/thrift/generator/server.ex
```diff
@@ -0,0 +1,139 @@
+defmodule Thrift.Generator.Server do
+  @moduledoc false
+
+  defmodule Framed do
```
pinterest-ebert: Modules should have a @moduledoc tag.

lib/thrift/generator/server.ex
```diff
@@ -0,0 +1,139 @@
...
+    alias Thrift.Parser.Models.Function
+    alias Thrift.Servers.BinaryFramed
+
+    def generate(service_module, service, file_group) do
+      functions = service.functions
+      |> Map.values
+      |> Enum.map(&generate_handler_function(file_group, service_module, &1))
+
+      quote do
+        defmodule Server.Framed do
```
pinterest-ebert: Modules should have a @moduledoc tag.

pinterest-ebert:

* 5 possible new issues (including those that may have been commented here).

You can see more details about this review at https://ebertapp.io/github/pinterest/elixir-thrift/pulls/116.

pinterest-ebert:

But beware that this branch is **[1 commit behind the `pinterest:thrift_tng` branch](https://github.com/pinterest/elixir-thrift/compare/scohen:framed_server...pinterest:thrift_tng)**, and a review of an up to date branch would produce more accurate results.

You can see more details about this review at https://ebertapp.io/github/pinterest/elixir-thrift/pulls/116.

lib/thrift/generator/behaviour.ex
```diff
@@ -7,6 +7,7 @@ defmodule Thrift.Generator.Behaviour do
   equivalent to their thrift counterparts.
   """
   alias Thrift.Parser.FileGroup
+  alias Thrift.Generator.Utils
```
jparise: Move this line above the other Parser line?

lib/thrift/generator/server.ex
```diff
@@ -0,0 +1,143 @@
...
+    end
+
+    def generate_handler_function(file_group, service_module, %Function{params: []} = function) do
+      fn_name = Atom.to_string(function.name)
+      handler_fn_name = Utils.underscore(function.name)
+      response_module = service_module
+      |> Module.concat(Service.service_module_name(function, :response))
+
+      handler = quote do
+          rsp = handler_module.unquote(handler_fn_name)()
```
jparise: This looks over-indented.

lib/thrift/generator/server.ex
```diff
@@ -0,0 +1,143 @@
...
+      end)
+
+      catchall = quote do
+        unhandled ->
+          {:server_error, %Thrift.TApplicationException{
+              message: "Server error: #{unhandled.message}",
+              type: Thrift.TApplicationException.exception_type(6)}}
+      end
+
+      rescue_blocks = List.flatten(rescue_blocks ++ [catchall])
```
jparise: I'm not sure if it would make a difference, but this is a little less work:

```elixir
List.flatten(rescue_blocks) ++ [catchall]
```

lib/thrift/generator/server.ex
```diff
@@ -0,0 +1,143 @@
...
+        rescue
+          unquote(rescue_blocks)
+        end
+      end
+    end
+
+    defp build_responder(:void, _) do
+      quote do
+        _ = rsp
+        :noreply
```
jparise: Indent.

lib/thrift/protocol/binary.ex
```diff
@@ -121,6 +131,38 @@ defmodule Thrift.Protocol.Binary do
     {:error, {:cant_decode_message, rest}}
   end
 
+  def deserialize(:application_exception, binary) when is_binary(binary) do
```
jparise: You'll need to update the `@spec deserialize` when you merge the upstream code. It's currently just looking for `:message_begin` as the first argument. We could just make that `atom` if there are going to be a variety of top-level deserialization types.

lib/thrift/servers/binary_framed/acceptor.ex
```diff
@@ -0,0 +1,73 @@
+defmodule Thrift.Servers.BinaryFramed.Acceptor do
+  @moduledoc """
```
jparise: Do you want this module to be part of the library's public API?

lib/thrift/servers/binary_framed/acceptor.ex
```diff
@@ -0,0 +1,73 @@
...
+
+  def handle_info(:accept, {socket, server_module, handler_module}) do
+    {:ok, r_sock} = :gen_tcp.accept(socket)
+
+    do_thrift_call({r_sock, server_module, handler_module})
+  end
+
+  defp do_thrift_call({socket, server_module, handler_module} = args) do
+    err = with {:ok, message}   <- :gen_tcp.recv(socket, 0),
+    parsed_response             <- Protocol.Binary.deserialize(:message_begin, message),
```
jparise: Indent a little? That would make this a bit easier to read to my eyes.

lib/thrift/servers/binary_framed/server.ex
```diff
@@ -0,0 +1,35 @@
+defmodule Thrift.Servers.BinaryFramed do
+  @moduledoc """
+  A server that uses the framed binary protocol.
+
+  Users should not use this module directly, and instead use one of the generated server modules.
```
jparise: Should we use `@moduledoc false` in that case?

lib/thrift/generator/server.ex
pguillory: Or:
```elixir
List.flatten(rescue_blocks, [catchall])
```

lib/thrift/generator/server.ex
```diff
@@ -0,0 +1,143 @@
+defmodule Thrift.Generator.Server do
+  @moduledoc false
+
+  defmodule Framed do
+    @moduledoc false
+
+    alias Thrift.Generator.{
```
dantswain: I dunno if you guys already have an established convention for this, but personally I prefer just listing out all of the aliases explicitly.  That makes it a lot easier to search for `Thrift.Generator.Service`, etc., and it makes it easier to refactor later on.

Disregard if you guys are already going with this style in other places.

lib/thrift/servers/binary_framed/acceptor.ex
```diff
@@ -0,0 +1,73 @@
...
+    GenServer.start_link(__MODULE__, [socket, server_module, handler_module])
+  end
+
+  def init([socket, server_module, handler_module]) do
+    Process.send_after(self, :accept, 1)
+    {:ok, {socket, server_module, handler_module}}
+  end
+
+  def handle_info(:accept, {socket, server_module, handler_module}) do
+    {:ok, r_sock} = :gen_tcp.accept(socket)
```
dantswain: I think this will block the entire GenServer until packets are received, which would lock out any kind of operational interaction.

lib/thrift/servers/binary_framed/acceptor.ex
```diff
@@ -0,0 +1,73 @@
...
+
+    do_thrift_call({r_sock, server_module, handler_module})
+  end
+
+  defp do_thrift_call({socket, server_module, handler_module} = args) do
+    err = with {:ok, message}   <- :gen_tcp.recv(socket, 0),
+    parsed_response             <- Protocol.Binary.deserialize(:message_begin, message),
+    {:ok, thrift_data}          <- handle_thrift_message(parsed_response, server_module, handler_module) do
+
+      :gen_tcp.send(socket, thrift_data)
```
dantswain: Is there a way to structure this so that the `gen_tcp` calls are within the same scope?  I find it easier to reason about code if paired side effects are kept close together - i.e.,
```
input = receive_input(...)
output = do_stuff_with(input)  # <- purely functional if possible
send_output(output)
```

test/generator/behaviour_test.exs
```diff
@@ -29,5 +30,6 @@ defmodule BehaviourTest do
     assert {:my_map, 1} in behaviour_specs
     assert {:my_map2, 1} in behaviour_specs
     assert {:struct_param, 1} in behaviour_specs
+    assert {:my_camel_cased_function, 1} in behaviour_specs
```
dantswain: Yay!

test/servers/binary_framed_test.exs
```diff
@@ -0,0 +1,166 @@
...
+
+      def server_exception do
+        raise "This wasn't supposed to happen"
+      end
+
+      def echo_struct(id_and_name), do: id_and_name
+
+      def returns_nothing, do: nil
+
+      def multiple_exceptions(exception_type) do
```
dantswain: Nit: these could be function clauses instead of a case.

lib/thrift/servers/binary_framed/acceptor.ex
```diff
@@ -0,0 +1,73 @@
...
+  use GenServer
+  require Logger
+
+  @spec start_link(pid, module, module) :: GenServer.on_start
+  def start_link(socket, server_module, handler_module) do
+    GenServer.start_link(__MODULE__, [socket, server_module, handler_module])
+  end
+
+  def init([socket, server_module, handler_module]) do
+    Process.send_after(self, :accept, 1)
```
pguillory: What's up with the send_after? Why 1ms? Why not send now?

test/servers/binary_framed_test.exs
```diff
@@ -0,0 +1,166 @@
...
+            {:DOWN, ^ref, _, _, _} ->
+              :ok
+          end
+      end
+    end
+
+    :rand.seed(:exs64, :erlang.now)
+
+
+    {:module, mod_name, _, _} = define_handler()
```
pguillory: No need to redefine the module for every test, right? It doesn't change, just define it once.

lib/thrift/servers/binary_framed/acceptor.ex
scohen: I don't think so;  I looked at several implementations, and this is, from what I can tell idiomatic.

You need to accept on the socket and then enter a loop, so I don't think you can have _all_ the gen_tcp calls in the same place. 3/4 of them are here. 

Maybe I just don't get what you want;

lib/thrift/generator/server.ex
scohen: I dunno, I've had it both ways and think that for > 3 aliases it saves enough to be worthwhile.

lib/thrift/generator/server.ex
```diff
@@ -0,0 +1,143 @@
+defmodule Thrift.Generator.Server do
+  @moduledoc false
+
+  defmodule Framed do
```
scohen: I can also save indentation by putting this in its own file / directory. 
you and @jparise need to duke this out; it's not fun going back and forth on this. Personally, I think BinaryFramedServer is redundant given it's in the Server.Framed module.

Either we make a module called Servers and have BinaryFramed in that or we keep the present nesting; I don't care which, but having both is redundant.

lib/thrift/generator/server.ex
scohen: That looks correctly indented to me.

test/servers/binary_framed_test.exs
```diff
@@ -0,0 +1,166 @@
...
+    {:ok, client} = Client.Framed.start_link("localhost", server_port, [])
+
+    {:ok, _} = Server.Framed.start_link(mod_name, server_port, [worker_count: 20])
```
pguillory: I'm not psyched about passing the handler module into the server's start_link. What's the use case for that, outside of this test? If we generate a ServerTest.Server, I think it would be reasonable to just expect the user to have defined a ServerTest.Handler and generate calls to ServerTest.Handler.ping or whatever.

lib/thrift/servers/binary_framed/acceptor.ex
scohen: Yes, yes it will. This is fine, as this is an acceptor process.

lib/thrift/servers/binary_framed/acceptor_supervisor.ex
```diff
@@ -0,0 +1,20 @@
...
+  """
+
+  alias Thrift.Servers.BinaryFramed.Acceptor
+  use Supervisor
+
+  def start_link(socket, server_module, handler_module, count) do
+    Supervisor.start_link(__MODULE__, [socket, server_module, handler_module, count])
+  end
+
+  def init([socket, server_module, handler_module, count]) do
```
scohen: and less explicit. And no one calls this manually.

lib/thrift/generator/server.ex
scohen: Neither suggestion works.

lib/thrift/servers/binary_framed/acceptor.ex
scohen: No. Moduledoc false?

pinterest-ebert:

But beware that this branch is **[2 commits behind the `pinterest:thrift_tng` branch](https://github.com/pinterest/elixir-thrift/compare/scohen:framed_server...pinterest:thrift_tng)**, and a review of an up to date branch would produce more accurate results.

You can see more details about this review at https://ebertapp.io/github/pinterest/elixir-thrift/pulls/116.

lib/thrift/generator/client/binary_framed.ex
```diff
@@ -0,0 +1,123 @@
...
+  alias Thrift.Parser.Models.Function
+  alias Thrift.Clients.BinaryFramed
+
+  def generate(service_module, service) do
+    functions = service.functions
+    |> Map.values
+    |> Enum.map(&generate_handler_function(service_module, &1))
+
+    quote do
+      defmodule Client.Framed do
```
pinterest-ebert: Modules should have a @moduledoc tag.

pinterest-ebert:

* 2 possible new issues (including those that may have been commented here).
* 2 fixed issues! 🎉

But beware that this branch is **[2 commits behind the `pinterest:thrift_tng` branch](https://github.com/pinterest/elixir-thrift/compare/scohen:framed_server...pinterest:thrift_tng)**, and a review of an up to date branch would produce more accurate results.

You can see more details about this review at https://ebertapp.io/github/pinterest/elixir-thrift/pulls/116.

test/servers/binary_framed_test.exs
scohen: That really limits your flexibility and screams "magic". Passing in handler modules is a fairly common idiom (used by ranch and the old erlang server). 

Also, it's helpful when testing to define a test server that differs from your prod server (I do this all the time, with zambezi, maia and electro); if you hard-code a module, you can't do that.

lib/thrift/generator/server/binary_framed.ex
```diff
@@ -0,0 +1,138 @@
...
+          |> unquote(response_module).BinaryProtocol.serialize()
+          {:reply, serialized_exception}
+      end
+    end)
+
+    catchall = quote do
+      unhandled ->
+        {:server_error, %Thrift.TApplicationException{
+            message: "Server error: #{unhandled.message}",
+            type: Thrift.TApplicationException.exception_type(6)}}
```
pguillory:
```
type: 6}}
```
Otherwise at runtime it's converting the number to an atom and then back to a number.

lib/thrift/servers/binary_framed/server.ex
```diff
@@ -0,0 +1,31 @@
...
+  @spec start_link(module, (1..65535), module, server_opts, GenServer.option) :: GenServer.on_start
+  def start_link(server_module, port, handler_module, opts, gen_server_opts) do
+    GenServer.start_link(__MODULE__, [server_module, port, handler_module, opts], gen_server_opts)
+  end
+
+  def init([server_module, port, handler_module, opts]) do
+    worker_count = Keyword.get(opts, :worker_count, 1)
+
+    {:ok, socket} = :gen_tcp.listen(port, [:binary, packet: 4, active: false])
+    {:ok, supervisor} = AcceptorSupervisor.start_link(socket, server_module, handler_module, worker_count)
```
pguillory: This supervisor is not part of a supervision tree, and it's linked to the server. If the server dies, it will kill all the connections as well. And when the whole application shuts down, it won't get a chance to close the connections cleanly (for instance by finishing the current request and not starting another). I wonder if we can set up a structure such that server process (holding the listen socket) and the connection processes (holding the regular sockets) are supervised independently. Maybe a topic for a future PR though.

lib/thrift/servers/binary_framed/acceptor_supervisor.ex
```diff
@@ -0,0 +1,20 @@
...
+    children = for id <- (1..count) do
+      worker_id = "#{inspect handler_module}_acceptor_#{id}"
```
fishcakez: `id` does not need to be an atom, I think `{handler, id}` would be a more natural data structure.

lib/thrift/servers/binary_framed/server.ex
fishcakez: With the current design I think this could be merged with `AcceptorSupervisor` as a single process/supervisor. This would ensure that lower down in the supervision tree (`AcceptorSupervisor` and its descendants) are terminated before the process in this position in the tree exits.

@pguillory here is an example where they are split: https://github.com/fishcakez/acceptor_pool/tree/master/examples/telnet_echo (requires OTP 19).

lib/thrift/generator/server/binary_framed.ex
```diff
@@ -0,0 +1,138 @@
...
+    end
+
+    rescue_blocks = List.flatten(rescue_blocks ++ [catchall])
+
+    quote do
+      try do
+        unquote(quoted_handler)
+      rescue
+      unquote(rescue_blocks)
```
fishcakez: What about throws and exits? Perhaps the catch all should use `catch`? However its possible for functions that raise an "unexpected" error or exit to leak messages or resources. Therefore if there is a catch all to reply on the socket the process would need to exit afterwards. This could be resolved by using a `Task.async` style process but it would be a performance hit. I think it would be preferable to close the socket on these `:server_error`s to prevent these issues without taking a performance hit. How will that affect clients? Is it possible for the server to inform clients that it is shutting down using the thrift protocol?

lib/thrift/servers/binary_framed/acceptor_supervisor.ex
```diff
@@ -0,0 +1,20 @@
...
+      worker(Acceptor, [socket, server_module, handler_module], id: worker_id)
+    end
+    supervise(children, strategy: :one_for_one)
```
fishcakez: The workers are permanent and the max restarts/max seconds will be the default 3/5. Given that the workers exit when their sockets close if a client machine goes down it will likely take down the server. I am not sure it is possible to have good fault tolerance with this simple design.

fishcakez:
A quick note on exceptions because this branch (as a whole) does not use them idiomatically.

The intention is to create exceptions using `exception/1`, which is defined inside exception module. Either with `MyException.exception(keyword)` or `raise MyException, keyword`. I think this would clean up and future proof exception handling code.

I think the use of integers is very hard to understand in the application exceptions type. Instead the atoms should be used directly and the integers hidden underneath. For example:
```elixir
MyException.exception([type: :bad_sequence_id])
```
or
```elixir
raise MyException, [message: msg, type: :bad_squence_id]
```
The `exception/1` callback can have an `@spec` for documentation and success typing.

Also there are functions in `Exception` than might be useful for formatting exceptions, rather than doing `exception.message` or similar. Perhaps you would want `Exception.format_banner`.

fishcakez:
> Also, I think you did this though there's a lot here and it's a little tricky to follow, but there should be an escape hatch if possible - i.e., allow someone to use this framework but handle the server aspect in their own way according to their own constraints.

Normally this is handled with a wrapper around `:gen_server.enter_loop` or a similar idea.

lib/thrift/generator/server/binary_framed.ex
scohen: Most clients have reconnect logic in them, so shutting down won't be a big deal; there's no way to deal with that at the protocol level.

lib/thrift/servers/binary_framed/acceptor.ex
```diff
@@ -0,0 +1,73 @@
...
+                message = Protocol.Binary.serialize(:message_begin, {:reply, sequence_id, name})
+
+                [message | <<0>>]
+            end
+
+    {:ok, reply}
+  end
+
+  def handle_thrift_message({:ok, {:oneway, _seq_id, name, args_binary}}, server_module, handler_module) do
+    spawn(server_module, :handle_thrift, [name, args_binary, handler_module])
```
scohen: I thought backpressure was handled by limiting the number of acceptors

lib/thrift/servers/binary_framed/acceptor_supervisor.ex
scohen: This is a really solid point. I'm still pretty terrible with supervision; and had a similar problem a while back. 
I don't know if we need _telecom grade_ fault tolerance in this server, but it should be tolerant of losing all its clients, as many of our clients only hold connections for 5 minutes.

lib/thrift/generator/server/binary_framed.ex
```diff
@@ -0,0 +1,138 @@
...
+    quote do
+      rsp = handler_module.unquote(handler_fn_name)(unquote_splicing(handler_args))
+      unquote(build_responder(function.return_type, response_module))
+    end
+    |> wrap_with_try_catch(function, file_group, response_module)
+  end
+
+  defp wrap_with_try_catch(quoted_handler, function, file_group, response_module) do
+    rescue_blocks = function.exceptions
+    |> Enum.map(fn exc ->
```
scohen: no, that doesn't work either. :-/

lib/thrift/servers/binary_framed/acceptor.ex
scohen: Ah, never mind; A client could fire off thousands of oneway requests agains our server this way. I see what you mean.

lib/thrift/servers/binary_framed/acceptor_supervisor.ex
scohen: @fishcakez, do you think it would make more sense to have a poolboy pool of acceptors and pick from there then?

lib/thrift/servers/binary_framed/server.ex
scohen: @pguillory super good point there; this would effectively kill things that don't (yet) need to be killed.

lib/thrift/servers/binary_framed/acceptor_supervisor.ex
scohen: or use your `acceptor_pool` library?

scohen: I didn't use ranch because this British guy told me that it had some issues. 

One of the immediate issues I see with ranch is that it doesn't allow for `packet: 4`, but I suppose I could handle that myself.

scohen: Ignore that last paragraph; I see that you can override those values, just not initially.

lib/thrift/generator/server/binary_framed.ex
scohen: That still didn't work, mainly because the exceptions i'm iterating over are exception metadata, not real exceptions. I've pushed up a different solution which gets rid of the List.flatten.

PR #118, opened 2016-12-31
Fix typedefs with container types

### LINKED ISSUE #111: Typedef set causes an error
The following thrift snippet:

```
  typedef set<i32> SetOfIntegers

  struct Thangs {
    1: optional SetOfIntegers magic_numbers
  }
```

Causes the following error on `mix thrift.generate`:

```
** (exit) an exception was raised:
    ** (CaseClauseError) no case clause matching: {:set, :i32}
        (thrift) lib/thrift/parser/resolver.ex:37: anonymous fn/2 in Thrift.Parser.Resolver.update/3
        (elixir) lib/map.ex:114: Map.do_new_transform/3
        (thrift) lib/thrift/parser/resolver.ex:36: Thrift.Parser.Resolver.update/3
        (elixir) lib/agent/server.ex:23: Agent.Server.handle_call/3
        (stdlib) gen_server.erl:615: :gen_server.try_handle_call/4
        (stdlib) gen_server.erl:647: :gen_server.handle_msg/5
        (stdlib) proc_lib.erl:247: :proc_lib.init_p_do_apply/3
```

scohen:
Fixes #111

### DIFF
--- a/lib/thrift/parser/resolver.ex
+++ b/lib/thrift/parser/resolver.ex
@@ -36,9 +36,7 @@ defmodule Thrift.Parser.Resolver do
     new_mappings = local_mappings
     |> Map.new(fn {name, val} ->
       case val do
-        val when is_tuple(val) ->
-          {:"#{include_name}.#{name}", val}
-        val when is_atom(val) ->
+        val when is_atom(val) or is_tuple(val) ->
           {:"#{include_name}.#{name}", val}
         val when is_map(val) ->
           {:"#{include_name}.#{name}", Map.put(val, :name, :"#{include_name}.#{name}")}
--- a/lib/thrift/parser/resolver.ex
+++ b/lib/thrift/parser/resolver.ex
@@ -36,6 +36,8 @@ defmodule Thrift.Parser.Resolver do
     new_mappings = local_mappings
     |> Map.new(fn {name, val} ->
       case val do
+        val when is_tuple(val) ->
+          {:"#{include_name}.#{name}", val}
         val when is_atom(val) ->
           {:"#{include_name}.#{name}", val}
         val when is_map(val) ->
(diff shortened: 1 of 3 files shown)


lib/thrift/parser/resolver.ex
```diff
@@ -36,6 +36,8 @@ defmodule Thrift.Parser.Resolver do
     new_mappings = local_mappings
     |> Map.new(fn {name, val} ->
       case val do
+        val when is_tuple(val) ->
+          {:"#{include_name}.#{name}", val}
         val when is_atom(val) ->
```
jparise: Could we also do `val when is_atom(val) or is_tuple(val)` and keep this one case?

test/generator/models_test.exs
```diff
@@ -149,4 +149,16 @@ defmodule Thrift.Generator.ModelsTest do
...
+  @thrift_file name: "complex_typedefs.thrift", contents: """
+  typedef set<i32> IntSet
+  struct Thangs {
+    1: optional IntSet numbers
+  }
+  """
+
+  thrift_test "it should be able to have typedefs" do
+    thang = %Thangs{numbers: MapSet.new([1, 2, 3])}
+    assert thang.numbers == MapSet.new([1, 2, 3])
```
dantswain: Would it be sufficient to `assert %Thangs{numbers: MapSet.new([1, 2, 3])}`?  The assertion here is really just asserting that struct access works.

scohen: this is asserting that the field exists. Yes, it's pedantic.

dantswain: If the field didn't exist then `%Thangs{numbers: MapSet.new([1, 2, 3])}` would fail, too, wouldn't it?

PR #119, opened 2016-12-31
Add missing :union data type

jparise:
Also, recognize :message_begin as a serialize/2 type.

### DIFF
--- a/lib/thrift.ex
+++ b/lib/thrift.ex
@@ -10,8 +10,8 @@ defmodule Thrift do
 
   @typedoc "Thrift data types"
   @type data_type ::
-    :bool | :byte | :i16 | :i32 | :i64 | :double |
-    :string | :struct | :union | :map | :set | :list
+    :bool | :byte | :i16 | :i32 | :i64 | :double | :string | :struct | :union |
+    {:map, data_type, data_type} | {:set, data_type} | {:list, data_type}
 
   @type i8   :: (-128..127)
   @type i16 :: (-32_768..32_767)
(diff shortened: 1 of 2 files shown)


jparise:
After looking at this more, I'm actually not sure if `:union` should be a top-level data type due to the way we map it to a `:struct` later on. If it does below at the top-level, should we also have a `Binary_type_id/1` function for it (that returns `@struct`)?

@pguillory or @scohen, what do you think?

pguillory:
That type_id spec is the language for telling `serialize/2` how to handle arbitrary inputs that aren't structs/unions/exceptions, which are the only things for which we generate serialization code. So if you have a list of strings you can say `serialize({:list, :string}, strings)` and that will work without having defined a list of strings as a type of field in some struct that got generated.

However, we don't have a generic deserialize function, and in fact maybe such a thing can't exist for structs/unions/exceptions, because we don't know what type of struct (or its field names) is represented by some bytes without having a corresponding IDL (i.e. a .thrift file). At best you could represent all generically deserialized Thrift structs as Elixir maps with integer keys instead of field names. Kinda like how the Wireshark plugin displays them, as I understand it, since it also lacks the IDL to completely interpret the bytes.

Basically it doesn't make a ton of sense to serialize structs/unions with this code. You only use this code if you don't have an IDL, and if you don't have an IDL you can't *deserialize* what you just serialized, so what's the point.

pguillory:
Also I think the spec might be wrong for maps/sets/lists. As in, `serialize` takes `{:list, recursive_type}` as its argument, not `:list`.

jparise:
@pguillory, my latest commit should clean that up.

PR #122, opened 2016-12-31
Remove the Apache Thrift Erlang runtime library

jparise:
We still have a copy under test/support/src/ that's used by our
integration tests, but we otherwise no longer have any need to ship a
copy of Apache Thrift Erlang runtime.


pguillory:
Fantastic.

PR #124, opened 2016-12-31
The Thrift compiler task now uses Thrift.Generator

jparise:
Our :thrift compiler is now implemented in terms of our Elixir code
generation framework. This drops our dependency on the Apache Thrift
compiler binary.

The new Thrift.Generator.targets/1 function is used to determine the
full set of generated output files for a given input .thrift file. This
ensures that we only generate Elixir source code for stale outputs.


jparise:
@dantswain Yes, the `thrift.generate` task is still there. We want the `compile.thrift` task because that naming convention makes it easy to include in Mix's compiler list:

```elixir
compilers: [:thrift | Mix.compilers]
```

The current compiler task reads configuration values from `Mix.Project.config`, which is also necessary for compiler-level integration. I can easily make it also read more command-line-based values, too, which would make `mix compile.thrift` as useful as our current `mix thrift.generate` task.

@pguillory, if that sounds good to you, I can make that change as well.

jparise:
@dantswain And for why there were two: before this revision, `compile.thrift` was tied to the `thrift` binary compiler while `thrift.generate` used our code generation. They are now mostly redundant.

pguillory:
@jparise OK by me.

PR #125, opened 2016-12-31
Require string for StructGenerator.generate label

pguillory:
Previously its argument type was inconsistent, being sometimes a string and
other times an atom.


scohen:
Wondering why you went with strings when atoms seem more idiomatic. NBD though.

PR #126, opened 2017-01-01
Bump our minimum Elixir version to 1.3

jparise:
This gives us the ability to use 1.3+ APIs as we develop our 2.0 branch.
Many other libraries (Absinthe, Dialyxir) also have a 1.3+ requirement.

Also, Elixir 1.4 is imminent, so add that to our testing matrix too.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,10 @@
 language: elixir
 elixir: 1.3.4
-otp_release: 19.2
+otp_release: 18.2
 matrix:
   include:
-    - elixir: 1.2.6
-      otp_release: 18.2
+    - elixir: 1.4.0-rc.1
+      otp_release: 19.2
 sudo: required
 dist: trusty
 services: docker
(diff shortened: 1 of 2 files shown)


pguillory:
1.3 is only about 6 months old. As a library user it can be frustrating to depend on something that forces you to live in the bleeding edge. (Granted this is a fairly new project and probably doesn't have a lot of users yet.) In general, how long do we want to support an Elixir version? Maybe a couple of years?

jparise:
I would argue that we are effectively writing a new library because all of the relevant APIs have yet to be published and we are labeling it "2.0" in the [semver](http://semver.org/)-sense. It should have no existing users and therefore no backwards compatibility or upgrade concerns.

I would further argue that now is the best time to set a higher Elixir version bar because it will be more difficult and less convenient to do so once this major version of our library is released. That would commit us to support Elixir 1.2 as our baseline for e.g. years.

pguillory:
@jparise agreed.

jparise:
One more bit of data from the recent Elixir Users Survey: https://www.dailydrip.com/blog/elixir-users-survey-2016-results

Over 90% of people surveyed are using Elixir 1.3. I'm not sure if that's representative of the user base at large, but it's one good data point.

PR #127, opened 2017-01-02
Enforce field requiredness during serialization

pguillory:
The [IDL spec][1] states that required fields must always be written during
serialization. There are also "default required" fields, i.e. fields specified
neither as required nor optional. Apparently they also should always be written,
though there are exceptions to which the spec alludes but about which it does
not deign to elaborate.

[1]: https://thrift.apache.org/docs/idl#field-requiredness

### DIFF
--- a/lib/thrift/generator/struct_binary_protocol.ex
+++ b/lib/thrift/generator/struct_binary_protocol.ex
@@ -769,25 +769,50 @@ defmodule Thrift.Generator.StructBinaryProtocol do
     end
   end
 
-  defp field_serializer(%Field{name: name, type: :bool, id: id}, _file_group) do
-    var = Macro.var(name, nil)
+  defp field_serializer(%Field{name: name, type: :bool, id: id, required: false}, _file_group) do
     quote do
-      case unquote(var) do
+      case unquote(Macro.var(name, nil)) do
         nil ->
           <<>>
         false ->
           <<unquote(@bool), unquote(id) :: size(16), 0>>
-        _ ->
+        true ->
           <<unquote(@bool), unquote(id) :: size(16), 1>>
+        _ ->
+          raise Thrift.MissingFieldException,
+                unquote("Optional boolean field #{inspect name} must be true, false, or nil")
+      end
+    end
+  end
+  defp field_serializer(%Field{name: name, type: :bool, id: id}, _file_group) do
+    quote do
+      case unquote(Macro.var(name, nil)) do
+        false ->
+          <<unquote(@bool), unquote(id) :: size(16), 0>>
+        true ->
+          <<unquote(@bool), unquote(id) :: size(16), 1>>
+        _ ->
+          raise Thrift.MissingFieldException,
+                unquote("Required boolean field #{inspect name} must be true or false")
+      end
+    end
+  end
+  defp field_serializer(%Field{name: name, required: false} = field, file_group) do
+    quote do
+      case unquote(Macro.var(name, nil)) do
+        nil ->
+          <<>>
+        _ ->
+          unquote(required_field_serializer(field, file_group))
       end
     end
   end
   defp field_serializer(%Field{name: name} = field, file_group) do
-    var = Macro.var(name, nil)
     quote do
-      case unquote(var) do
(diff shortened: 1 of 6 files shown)


dantswain:
@pguillory were you able to benchmark this?  It doesn't look like it should have much impact but I'm maybe a little over-skeptical because I know that this validation makes the Ruby client significantly slower.

lib/thrift/generator/struct_binary_protocol.ex
```diff
@@ -769,30 +769,55 @@ defmodule Thrift.Generator.StructBinaryProtocol do
...
+      case unquote(Macro.var(name, nil)) do
         nil ->
           <<>>
         false ->
           <<unquote(@bool), unquote(id) :: size(16), 0>>
+        true ->
+          <<unquote(@bool), unquote(id) :: size(16), 1>>
         _ ->
+          raise Thrift.MissingFieldException,
+                unquote("Optional boolean field #{inspect name} must be true, false, or nil")
```
jparise: It would be helpful to also include the struct/module name in addition to the field name, if possible. Below as well.

lib/thrift/generator/struct_binary_protocol.ex
```diff
@@ -769,30 +769,55 @@ defmodule Thrift.Generator.StructBinaryProtocol do
...
         nil ->
           <<>>
         _ ->
           unquote(required_field_serializer(field, file_group))
       end
     end
   end
+  defp field_serializer(%Field{name: name} = field, file_group) do
+    quote do
+      case unquote(Macro.var(name, nil)) do
```
jparise: I might consider using a `if` macro here instead of the `case`. I find them both readable but the `if` would be shorter and the exception `else` clause more obvious.

dantswain:
One way to support "requiredness" is to do so optionally via a function. 

```
%MyStruct{}
|> MyStruct.validate!
|> MyStruct.serialize!
```

That way users could provide their own validation without having to repeat the same operations, or enforce validation upstream and skip it here for performance reasons.

Though if the performance hit for this is negligible then it's not a big deal.

jparise:
> That way users could provide their own validation without having to repeat the same operations, or enforce validation upstream and skip it here for performance reasons.

I believe this is how the Python library works, too.

lib/thrift/generator/struct_binary_protocol.ex
```diff
@@ -769,30 +769,55 @@ defmodule Thrift.Generator.StructBinaryProtocol do
...
-      case unquote(var) do
+      case unquote(Macro.var(name, nil)) do
         nil ->
           <<>>
         false ->
           <<unquote(@bool), unquote(id) :: size(16), 0>>
+        true ->
+          <<unquote(@bool), unquote(id) :: size(16), 1>>
         _ ->
+          raise Thrift.MissingFieldException,
```
scohen: Warning: Pedantry.
We're raising a MissingFieldException, but the field isn't technically missing; it's there, but with an invalid field. Consider an `InvalidFieldValueException` instead?

pguillory: It's a good point. We may add validation on other things that are not fields. (Like what if your list of ints contains a string.) How about `InvalidValueException`?

lib/thrift/generator/struct_binary_protocol.ex
pguillory: `if` takes the `else` branch on nil or false, but we only want to test for nil here.

pguillory:
I ran the benchmark before and after the change, not seeing a significant difference. The numbers aren't consistent enough between runs to see differences less than a few percent though. Anyway the serialization+iolist_size benchmark is about 860 to 870 µs/op on my laptop both before and after.

Required fields are pretty insidious. If we make validation optional and one system erroneously omitted the field, causing another to be unable to parse a message, that's a frustrating situation. For that reason if we make validation optional, it should default to true. But I don't really think it should be optional. If the IDL says a field is a bool, it better not be a string. If it says a field is required, it better be there.

I'm in favor of adding extra implicit validation, in fact, where it doesn't significantly impact performance. For instance we could validate that integers do not exceed their ranges.

Issue #112, opened 2016-12-30
Generator should specify which file causes an error


dantswain:
Running `mix thrift.generate` with a directory that has a bunch of thrift files in it, I get this error:

```
** (MatchError) no match of right hand side value: {:error, {32, :thrift_parser, ['syntax error before: ', '\']\'']}}
    lib/thrift/parser.ex:19: Thrift.Parser.parse/1
    lib/thrift/parser/parsed_file.ex:13: Thrift.Parser.ParsedFile.new/1
    lib/thrift/parser.ex:57: Thrift.Parser.parse_file/1
    lib/thrift/generator.ex:11: Thrift.Generator.generate!/2
    lib/mix/tasks/thrift.generate.ex:32: anonymous fn/3 in Mix.Tasks.Thrift.Generate.run/1
    (elixir) lib/enum.ex:1623: Enum."-reduce/3-lists^foldl/2-0-"/3
    lib/mix/tasks/thrift.generate.ex:31: Mix.Tasks.Thrift.Generate.run/1
    (mix) lib/mix/task.ex:296: Mix.Task.run_task/3
```

I don't think any of my files have thrift syntax errors, which is another issue that I'll try to track down, but this doesn't tell me which file is causing the error.

dantswain:
Closed by #128

PR #128, opened 2017-01-02
Raise a meaningful exception on parse errors

dantswain:
Re #112 - raises an exception whose message includes the filename and, if possible, the line number.

To be idiomatic, there are a handful of functions that should probably be banged - e.g., `Parser.parse!` since it can raise an exception - but I wasn't sure how intrusive that would be.  If you're ok with it I can make that change.

It felt a little clunky to write a temp file, but this doesn't work with the existing `thrift_test` framework because the exception would halt the other tests.


lib/thrift/exceptions.ex
```diff
@@ -22,3 +22,24 @@ defmodule Thrift.Union.TooManyFieldsSetException do
...
 end
+
+defmodule Thrift.FileParseException do
+  @moduledoc """
+  This exception occurs when a thrift file fails to parse
+  """
+
+  defexception message: nil
+
+  def exception({file_ref, error}) do
```
jparise: Add a `@spec` here?

test/parser_test.exs
```diff
@@ -697,4 +699,43 @@ defmodule ParserTest do
...
+
+    # should raise an error on the included file,
+    # since that is where the syntax error is
+    assert_raise(
+      Thrift.FileParseException,
+      ~r/#{path} on line 1:/,
+      fn -> parse_file(other_path) end
+    )
+
+    File.rm_rf!(dir)
```
jparise: We won't get here if the assertion above fails.

dantswain: True.  I didn't like the way this was set up, but wasn't sure where else to put it.  It seemed awkward to have a setup/teardown execute this for every test in the file where this is the only  test that actually uses that directory.  I could move it to its own file, I suppose?

dantswain:
@jparise good question.  I more or less just picked this option because I assumed the exceptions file was where you guys _wanted_ exceptions.  I'm happy to move it around.

I'm not sure why ebert keeps failing on this - has that been happening with other PRs?

jparise:
> I'm not sure why ebert keeps failing on this - has that been happening with other PRs?

They seem to be having some sort of outage. I reported it yesterday. We can just ignore that failure until they're back online.

dantswain:
Does anyone know how to force a lexer error?  I want to write a test to handle the lexer error case but I don't grok how the lexer/parser work together and all I can get is parser errors.

dantswain:
Found a way to get a lexer error ✅

dantswain:
@jparise I'm fine with that - `Thrift.Parser.FailedFileException`?  Sounds a little awkward but `Thrift.Parser.Exception` doesn't seem specific enough.  I'm open to suggestions.

test/parser/parse_error_test.exs
```diff
@@ -0,0 +1,85 @@
...
+
+  setup do
+    File.rm_rf!(@test_file_dir)
+    File.mkdir_p!(@test_file_dir)
+    on_exit fn -> File.rm_rf!(@test_file_dir) end
+  end
+
+  test "a file that throws parser errors raises an exception" do
+    contents = """
+    stract Typo {
```
scohen: I keep reading this with some kind of southern accent.

jparise:
> Thrift.Parser.FailedFileException? Sounds a little awkward but Thrift.Parser.Exception doesn't seem specific enough. I'm open to suggestions.

`Thirft.Parser.ParseException` feels good to me.

lib/thrift/exceptions.ex
```diff
@@ -22,3 +22,29 @@ defmodule Thrift.Union.TooManyFieldsSetException do
...
+
+  # display the line number if we get it
+  defp format_error({line_no, :thrift_parser, errors}) do
+    "on line #{line_no}: #{inspect errors}"
+  end
+  defp format_error({{line_no, :thrift_lexer, errors}, _}) do
+    "on line #{line_no}: #{inspect errors}"
+  end
+  defp format_error(error) do
+    ":#{inspect error}"
```
pguillory: This message won't be formatted quite right. There will be a space before the `:` instead of after.

The errors look suspiciously like iolists. I wonder if that is consistently true. Would be a bit nicer to display `syntax error before: "stract"` than `['syntax error before: ', ['"stract"']]`.

scohen: Now they make sense!

dantswain: I think that's true of the parser errors but not the lexer.  I changed the parser error case.

PR #129, opened 2017-01-02
Minor improvements to the Thrift compile task

jparise:
- Our command line options now more closely mimic the Apache Thrift
  compiler's. We support both long and "short" option formats.
- The compile task's output now simply prints "Compiling N files", which
  is the Elixir 1.3+ output convention. Use the `--verbose` command line
  option to see per-file output.
- The compiler task's unit test helper code has been simplified a bit.

### DIFF
--- a/lib/mix/tasks/compile.thrift.ex
+++ b/lib/mix/tasks/compile.thrift.ex
@@ -21,7 +21,7 @@ defmodule Mix.Tasks.Compile.Thrift do
   ## Command line options
 
     * `--force` - forces compilation regardless of modification times
-    * `--output` - set the output directory, overriding the `:thrift_output`
+    * `--out` - set the output directory, overriding the `:thrift_output`
       configuration value
 
   ## Configuration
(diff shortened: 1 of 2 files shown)


lib/mix/tasks/compile.thrift.ex
```diff
@@ -21,7 +21,7 @@ defmodule Mix.Tasks.Compile.Thrift do
   ## Command line options
 
     * `--force` - forces compilation regardless of modification times
-    * `--output` - set the output directory, overriding the `:thrift_output`
+    * `--out` - set the output directory, overriding the `:thrift_output`
```
pguillory: We could include the short option forms in the docs. Also the verbose option isn't mentioned.

lib/mix/tasks/compile.thrift.ex
```diff
@@ -78,8 +79,8 @@ defmodule Mix.Tasks.Compile.Thrift do
     Enum.empty?(targets) || Mix.Utils.stale?([thrift_file], targets)
   end
 
-  defp generate(%FileGroup{initial_file: thrift_file} = group, output_dir) do
+  defp generate(%FileGroup{} = group, output_dir, opts) do
     Thrift.Generator.generate!(group, output_dir)
-    Mix.shell.info "Compiled #{thrift_file}"
+    if opts[:verbose], do: Mix.shell.info "Compiled #{group.initial_file}"
```
pguillory: I find the one-line `if` form a bit harder to read than:
```elixir
if opts[:verbose] do
  Mix.shell.info...
end
```

PR #130, opened 2017-01-02
Enum and union behaviours

scohen:
We were not generating behaviours for enum and union fields.

### DIFF
--- a/lib/thrift/generator/behaviour.ex
+++ b/lib/thrift/generator/behaviour.ex
@@ -9,6 +9,7 @@ defmodule Thrift.Generator.Behaviour do
   alias Thrift.Generator.Utils
   alias Thrift.Parser.FileGroup
   alias Thrift.Parser.Models.{
+    Exception,
     Field,
     Struct,
     StructRef,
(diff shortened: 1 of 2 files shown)


lib/thrift/generator/behaviour.ex
```diff
@@ -65,6 +67,17 @@ defmodule Thrift.Generator.Behaviour do
     |> FileGroup.resolve(ref)
     |> typespec(file_group)
   end
+  defp typespec(%TEnum{} = enum, _) do
```
jparise: Is `enum` unused hre?

lib/thrift/generator/behaviour.ex
```diff
@@ -65,6 +67,17 @@ defmodule Thrift.Generator.Behaviour do
...
+    quote do
+      pos_integer
+    end
+  end
+  defp typespec(%Union{name: name}, file_group) do
+    dest_module = FileGroup.dest_module(file_group, name)
+    quote do
+      %unquote(dest_module){}
+    end
+  end
```
pguillory: Do we also need one for Exception, while you're at it?

test/generator/behaviour_test.exs
```diff
@@ -31,5 +48,7 @@ defmodule BehaviourTest do
     assert {:my_map2, 1} in behaviour_specs
     assert {:struct_param, 1} in behaviour_specs
     assert {:my_camel_cased_function, 1} in behaviour_specs
+    assert {:get_state, 0} in behaviour_specs
+    assert {:get_sor_t, 0} in behaviour_specs
```
pguillory: getSOrT would be the camelcase corresponding to get_s_or_t.

lib/thrift/generator/behaviour.ex
```diff
@@ -65,6 +67,17 @@ defmodule Thrift.Generator.Behaviour do
     |> FileGroup.resolve(ref)
     |> typespec(file_group)
   end
+  defp typespec(%TEnum{} = enum, _) do
+    quote do
+      pos_integer
```
pguillory: non_neg_integer according to the spec.

lib/thrift/generator/behaviour.ex
scohen: Ok, but that'd be very weird!

PR #132, opened 2017-01-03
Small improvements to the binary protocol type IDs

jparise:
1. Fix the `type_id` typespec to use the correct value range.
2. Add a formal `@stop` type identifier (0) and use it below.
3. Serialize the `@stop` byte as a signed value, [per the spec](https://erikvanoosten.github.io/thrift-missing-specification/#_struct_encoding). (This makes
   no difference in practice, but it's more correct.)

### DIFF
--- a/lib/thrift/protocol/binary.ex
+++ b/lib/thrift/protocol/binary.ex
@@ -11,20 +11,21 @@ defmodule Thrift.Protocol.Binary do
   alias Thrift.TApplicationException
 
   @typedoc "Binary protocol field type identifier"
-  @type type_id :: (2..15)
+  @type type_id :: 0 | 2 | 3 | 4 | 6 | 8 | 10 | 11 | 12 | 13 | 14 | 15
   @type deserializable :: :message_begin | :application_exception
 
-  @bool 2
-  @byte 3
+  @stop   0
+  @bool   2
+  @byte   3   # also: i8
   @double 4
-  @i16 6
-  @i32 8
-  @i64 10
+  @i16    6
+  @i32    8
+  @i64    10
   @string 11
   @struct 12
-  @map 13
-  @set 14
-  @list 15
+  @map    13
+  @set    14
+  @list   15
 
   @spec type_id(Thrift.data_type) :: type_id
   defp type_id(:bool),      do: @bool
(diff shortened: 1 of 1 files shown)


lib/thrift/protocol/binary.ex
```diff
@@ -105,7 +106,7 @@ defmodule Thrift.Protocol.Binary do
   end
   def serialize(:application_exception, %TApplicationException{message: message, type: type}) do
     <<@string::size(8), 1::16-signed, byte_size(message)::size(32), message::binary,
-      @i32::size(8), 2::16-signed, type::32-signed, 0::size(8)>>
+      @i32::size(8), 2::16-signed, type::32-signed, @stop::8-signed>>
```
pguillory: I like `@stop` as a hint to the reader. Why make it signed though? It's reasonable to expect the reader to understand that elements in a binary expression are unsigned bytes unless otherwise annotated. You see it in `iex` all the time, for instance. So I'd just end with `@stop>>` here. Probably drop the `::size(8)` where we use it, too.

lib/thrift/protocol/binary.ex
```diff
@@ -11,20 +11,21 @@ defmodule Thrift.Protocol.Binary do
   alias Thrift.TApplicationException
 
   @typedoc "Binary protocol field type identifier"
-  @type type_id :: (2..15)
+  @type type_id :: 0 | 2 | 3 | 4 | 6 | 8 | 10 | 11 | 12 | 13 | 14 | 15
```
pguillory: I like enumerating the specific values.

Zero doesn't belong in here though. A stop byte isn't a field type and isn't used in the same way. For example, it's not a valid argument to `skip_field/1`.

jparise: 0/STOP is considered a field type in the Apache Thrift implementation, but I agree that it isn't useful to include in this list.

lib/thrift/protocol/binary.ex
jparise: It's always read and written as an `i8` (the same as a field type), but as it's always going to be 0 anyway, I like the idea of just writing `@stop` for the sake of brevity.

PR #133, opened 2017-01-03
Always treat struct field IDs as ::16-signed

jparise:
We were using using `::size(16)` in many cases, which is an unsigned
integer value. The [specification](https://erikvanoosten.github.io/thrift-missing-specification/#_struct_encoding) says this should be a signed value.


jparise:
This is unlikely to mean much in practice, but it's good to follow the spec consistently.

Issue #136, opened 2017-01-03
Validation is incorrect for "Default requiredness" fields


scohen:
Given the following thrift spec:
```thrift
struct Thang {
  1: required i64 thang_id,
  2: string thang_name
}
```
The field `thang_name` exhibits "Default requiredness" and does not have to be present. However, our validator blows up and demands that this field be set. 
 
The correct behavior is to allow this field to remain unset, and not serialize it.

The section about Requiredness here http://thrift.apache.org/docs/idl  is helpful

PR #137, opened 2017-01-03
Allow 'default required' fields to be nil

pguillory:
We discovered that other Thrift implementations we use internally do not
serialize unset 'default required' fields. This change brings our
implementation's behavior in line with the others.


pguillory:
Fixes #136.

scohen:
to @dantswain's point, we should probably change some of the old tests to not specify optional and make use of default requiredness.

PR #139, opened 2017-01-03
Server improvements

scohen:
The server's start_link didn't actually start_link. I've changed this to
bring the server's supervision into the user's application.

Also, we weren't properly cleaning up after server errors, so now I just crash.

@fishcakez, thanks for pointing all of this out.

### DIFF
--- a/lib/thrift/generator/server/binary_framed.ex
+++ b/lib/thrift/generator/server/binary_framed.ex
@@ -16,6 +16,7 @@ defmodule Thrift.Generator.Server.BinaryFramed do
     quote do
       defmodule Server.Framed do
         @moduledoc false
+        require Logger
 
         alias Thrift.Servers.BinaryFramed
         defdelegate stop(name), to: BinaryFramed
@@ -94,6 +95,7 @@ defmodule Thrift.Generator.Server.BinaryFramed do
       Thrift.TApplicationException ->
         quote do
           unhandled ->
+            Logger.error("Unhandled exception: #{Exception.message(unhandled)}")
             {:server_error, Thrift.TApplicationException.exception(
               message: "Server error: #{Exception.message(unhandled)}",
               type: :internal_error)}
--- a/lib/thrift/generator/server/binary_framed.ex
+++ b/lib/thrift/generator/server/binary_framed.ex
@@ -95,7 +95,7 @@ defmodule Thrift.Generator.Server.BinaryFramed do
       Thrift.TApplicationException ->
         quote do
           unhandled ->
-            Logger.error("Unhandled exception: #{Exception.message(unhandled)}")
+            Logger.error("Exception not defined in thrift spec was thrown: #{Exception.message(unhandled)}")
             {:server_error, Thrift.TApplicationException.exception(
               message: "Server error: #{Exception.message(unhandled)}",
               type: :internal_error)}
--- a/lib/thrift/generator/server/binary_framed.ex
+++ b/lib/thrift/generator/server/binary_framed.ex
@@ -90,17 +90,8 @@ defmodule Thrift.Generator.Server.BinaryFramed do
   end
 
   defp wrap_with_try_catch(quoted_handler, function, file_group, response_module) do
-    rescue_blocks = function.exceptions ++ [Thrift.TApplicationException]
+    rescue_blocks = function.exceptions
     |> Enum.flat_map(fn
-      Thrift.TApplicationException ->
-        quote do
-          unhandled ->
-            Logger.error("Exception not defined in thrift spec was thrown: #{Exception.message(unhandled)}")
-            {:server_error, Thrift.TApplicationException.exception(
-              message: "Server error: #{Exception.message(unhandled)}",
-              type: :internal_error)}
-        end
-
       exc ->
         resolved = FileGroup.resolve(file_group, exc)
         dest_module = FileGroup.dest_module(file_group, resolved.type)
@@ -119,7 +110,13 @@ defmodule Thrift.Generator.Server.BinaryFramed do
       try do
         unquote(quoted_handler)
       rescue
-      unquote(rescue_blocks)
+        unquote(rescue_blocks)
+      catch kind, reason ->
+        formatted_exception = Exception.format(kind, reason, System.stacktrace)
+        Logger.error("Exception not defined in thrift spec was thrown: #{formatted_exception}")
+        {:server_error, Thrift.TApplicationException.exception(
+          message: "Server error: #{formatted_exception}",
+          type: :internal_error)}
       end
     end
   end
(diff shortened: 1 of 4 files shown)


lib/thrift/generator/server/binary_framed.ex
```diff
@@ -94,6 +95,7 @@ defmodule Thrift.Generator.Server.BinaryFramed do
       Thrift.TApplicationException ->
         quote do
           unhandled ->
+            Logger.error("Unhandled exception: #{Exception.message(unhandled)}")
```
dantswain: Should there be some context hint in this?  `"Unhandled TApplicationException"` or something?

lib/thrift/servers/binary_framed/server.ex
```diff
@@ -3,24 +3,31 @@ defmodule Thrift.Servers.BinaryFramed do
 
   @type server_opts :: [
     worker_count: pos_integer,
-    name: atom
+    name: atom,
+    max_restarts: non_neg_integer,
```
dantswain: :100: Thank you... so many people don't do this.

lib/thrift/generator/server/binary_framed.ex
scohen: Well, it's not really a TApplicationException, it's an exception you didn't handle properly.

"Exception was thrown but not declared in your thrift spec?"

test/servers/binary_framed_test.exs
```diff
@@ -75,11 +75,13 @@ defmodule Servers.BinaryFramedIntegrationTest do
...
 
   def stop_server(server_pid) do
-    ref = Process.monitor(server_pid)
-    Server.Framed.stop(server_pid)
-    receive do
-      {:DOWN, ^ref, _, _, _} ->
-        :ok
+    if Process.alive?(server_pid) do
+      ref = Process.monitor(server_pid)
+      Server.Framed.stop(server_pid)
```
pguillory: There's a race condition. The process could be alive when we call `Process.alive?` but dead when we get to the `Server.Framed.stop`. You could catch the noproc though.
```elixir
ref = Process.monitor(server_pid)
try do
  Server.Framed.stop(server_pid)
catch :exit, :noproc -> :ok
end
assert_receive {:DOWN, ^ref, _, _, _}, 5000
```

lib/thrift/generator/server/binary_framed.ex
```diff
@@ -95,7 +95,7 @@ defmodule Thrift.Generator.Server.BinaryFramed do
       Thrift.TApplicationException ->
         quote do
           unhandled ->
-            Logger.error("Unhandled exception: #{Exception.message(unhandled)}")
+            Logger.error("Exception not defined in thrift spec was thrown: #{Exception.message(unhandled)}")
```
dantswain: Works for me.  Just something to clue the user in on where the exception came from.

lib/thrift/generator/server/binary_framed.ex
pguillory: This just logs the message, right? I think we could get the stack trace too with Exception.format.

lib/thrift/generator/server/binary_framed.ex
fishcakez: `rescue` does not rescue all exceptions, only error exceptions. To rescue all exceptions you require `catch`. I think the current behaviour hides too much information. I think this should be written as:

```elixir
catch
  kind, reason ->
    stack = System.stacktrace()
end
```
The TApplicationException could the use `Exception.format(kind, reason, stack)` instead of `Exception.message/1`. This will show all the information. Often the `stack` is the most useful piece of information.

It would also be good to do `:erlang.raise(kind, reason, stack)` instead of the `{:shutdown, :server_error}` so that link'ed process gets the same exit reason as would occur if the exception was not caught. The `:shutdown` tuple reason could still cause similar issues to the `:transient` problem we discussed when a user accidentally terminates a transient child.

An alternative would be to do the equivalent of
```elixir
catch
  :exit, reason -> {:exit, reason}
  :error, reason -> {:exit, {reason, System.stacktrace()}}
  :throw, reason -> {:exit, {{:nocatch, reason}, System.stacktrace()}}
end
```

Then call exit({exit_reason, {handler_module, :function, args}}). The `Exception` module will format this really nicely with `Exception.format_exit/1` because it takes the form `{exit_reason, {mod, fun, args}}`. The tricky with the `catch` produces the equivalent `exit_reason` to if the process had not caught the exception. We use this heavily in elixir-* projects (especially elixir itself).

I am not sure which is better. An OTP behaviour would usually create the same `exit_reason` (but without the MFA) and exit with that. It would also log a detailed error message that contains the equivalent information to including the mod-fun-args, and maybe some other info too.

test/servers/binary_framed_test.exs
fishcakez: Its a good catch but the `catch` does not have the correct format, it would be as the mod-fun-args in https://github.com/pinterest/elixir-thrift/pull/139#discussion_r94472017: `:exit, {:noproc, {GenServer, :call, _}}`. You would also require to catch `:shutdown` because ExUnit process will exit with reason `:shutdown` and that will propagate to the thrift server.

Actually the thrift server will be linked to the test process and trapping exits. This means that it will always have the `:EXIT` message in its message queue before the `:system` call message made by `Server.Framed.stop/1` because link signals are sent before monitor signals - and on_exit is triggered by a monitor. Therefore it should just be a case of monitoring and waiting for any `:DOWN`.

Also note that when using `GenServer.stop` et al a monitor is not required because the call will monitor the process and wait for it a `:DOWN` before returning.

test/servers/binary_framed_test.exs
```diff
@@ -143,9 +137,8 @@ defmodule Servers.BinaryFramedIntegrationTest do
   end
 
   thrift_test "it can handle unexpected exceptions", ctx do
-    expected_exception = %TAE{message: "Server error: This wasn't supposed to happen",
-                              type: :internal_error}
-    assert {:error, {:exception, expected_exception}} == Client.Framed.server_exception(ctx.client)
+    assert {:error, {:exception, %TAE{message: msg, type: :internal_error}}} = Client.Framed.server_exception(ctx.client)
+    assert String.contains?(msg, "Server error: ** (RuntimeError) This wasn't supposed to happen")
```
jparise: I prefer a simpler regex match for this:

```elixir
assert msg =~ "Server error: ** (RuntimeError) This wasn't supposed to happen"
```

test/servers/binary_framed_test.exs
fishcakez: @pguillory sorry you had the correct catch pattern as `GenServer.stop` has special handling for `:timeout` and `:noproc`.

PR #146, opened 2017-01-04
Proofread README

pguillory:
- Fixed a few typos.
- Fixed output module paths.
- Change "IO list" references to "iodata" as seems to be preferred by Elixir docs.


scohen:
This is all going to change when #148 is merged.

pguillory:
True. Can we merge this and then update the README in #148?

PR #147, opened 2017-01-04
Generate const default values

### LINKED ISSUE #140: Const default values fail to generate
Example thrift:

```
  const string DefaultRootPassword = "abc123"

  struct BestPracticeSecurity {
    1: optional string root_password = DefaultRootPassword
  }
```

Results in

```
** (MatchError) no match of right hand side value: [:DefaultRootPassword]
    (thrift) lib/thrift/parser/models.ex:105: Thrift.Parser.Models.TEnumValue.new/1
    (thrift) src/thrift_parser.yrl:132: :thrift_parser.yeccpars2_76/7
    (thrift) /usr/local/Cellar/erlang/19.1/lib/erlang/lib/parsetools-2.1.3/include/yeccpre.hrl:57: :thrift_parser.yeccpars0/5
    (thrift) lib/thrift/parser/parsed_file.ex:12: Thrift.Parser.ParsedFile.new/1
    (thrift) lib/thrift/parser.ex:61: Thrift.Parser.parse_file/1
    (thrift) lib/thrift/generator.ex:39: Thrift.Generator.generate!/2
    (elixir) lib/enum.ex:925: anonymous fn/3 in Enum.flat_map/2
    (elixir) lib/enum.ex:1623: Enum."-reduce/3-lists^foldl/2-0-"/3
    (elixir) lib/enum.ex:924: Enum.flat_map/2
    (thrift) test/support/lib/thrift_test_case.ex:47: ThriftTestCase.compile_and_build_erlang_helpers/2
    (thrift) expanding macro: ThriftTestCase.__before_compile__/1
    test/generator/models_test.exs:1: Thrift.Generator.ModelsTest (module)
    (elixir) lib/code.ex:363: Code.require_file/2
    (elixir) lib/kernel/parallel_require.ex:57: anonymous fn/2 in Kernel.ParallelRequire.spawn_requires/5

```

pguillory:
Previously when the parser found a reference where it expected a static value,
it treated it as an enum. However it could also be a const. Now the parser
turns any reference into a StructRef and leaves it to the next layer to resolve
the reference.

Fixes #140.

### DIFF
--- a/test/generator/binary_protocol_test.exs
+++ b/test/generator/binary_protocol_test.exs
@@ -397,4 +397,51 @@ defmodule Thrift.Generator.BinaryProtocolTest do
     assert_serializes %OptionalField{},                 <<0>>
     assert_serializes %OptionalField{val: 123},         <<3, 0, 1, 123, 0>>
   end
+
+  @thrift_file name: "const.thrift", contents: """
+  struct ConstStructVal {
+    1: byte num
+  }
+
+  const bool ConstBool = true
+  const byte ConstByte = 5
+  const double ConstDouble = 5.0
+  const i16 ConstI16 = 5
+  const i32 ConstI32 = 5
+  const i64 ConstI64 = 5
+  const string ConstString = "abc123"
+  const ConstStructVal ConstStruct = {"num": 5}
+  const map<string, byte> ConstMap = {"a": 1, "b": 2}
+  const set<string> ConstSet = ["a", "b"]
+  const list<string> ConstList = ["a", "b"]
+
+  struct ConstStruct {
+    1: bool bool_val = ConstBool,
+    2: byte byte_val = ConstByte,
+    3: double double_val = ConstDouble,
+    4: i16 i16_val = ConstI16,
+    5: i32 i32_val = ConstI32,
+    6: i64 i64_val = ConstI64,
+    7: string string_val = ConstString,
+    # 8: ConstStructVal struct_val = ConstStruct,
+    13: map<string, byte> map_val = ConstMap,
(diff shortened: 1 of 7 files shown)


pguillory:
Yeah we should rename StructRef in a subsequent PR. It's a reference to anything at this point. Maybe just `Ref`?

Another option would be to disambiguate between types and values, i.e. `TypeRef` and `ValueRef`. We know at parse time which one to expect, and a function would only really use one or the other. The newly added `default_value` only deals with value references, and the serialization generators only deal with type references. I think only the resolver would need to support both. This approach would help us catch things like:
```elixir
const i32 Num = 5
typedef list<Num> Nums   # oops, Num is not a type
```

jparise:
@pguillory, 👍  to distinct `TypeRef` and `ValueRef` types.

test/generator/binary_protocol_test.exs
```diff
@@ -397,4 +397,51 @@ defmodule Thrift.Generator.BinaryProtocolTest do
...
+    6: i64 i64_val = ConstI64,
+    7: string string_val = ConstString,
+    # 8: ConstStructVal struct_val = ConstStruct,
+    13: map<string, byte> map_val = ConstMap,
+    14: set<string> set_val = ConstSet,
+    15: list<string> list_val = ConstList,
+  }
+  """
+
+  thrift_test "test..." do
```
dantswain: "test..."?

dantswain:
:tada: Thanks @pguillory !

test/generator/binary_protocol_test.exs
pguillory: Whoops!

PR #156, opened 2017-01-05
Implement deserialization of maps/sets/lists of binaries

pguillory:
Seems to have been overlooked. We implemented strings, which are effectively
the same in Elixir, but didn't check specifically for a type of :binary in
every situation.


scohen:
@pguillory yes, that's what I thought. NBD.

PR #162, opened 2017-01-06
Readme faq

scohen:
Added a small FAQ.


README.md
```diff
@@ -287,3 +287,54 @@ You can use these features to support additional languages, protocols, and serve
...
+On the other hand, this implementation uses Elixir to write Elixir that's specific
+ to _your_ thrift data. This serialization logic is then compiled, and that compiled code
+is what converts your data into bytes. We've spent a lot of time making sure
+that the generated serialization code takes advantage of several of the optimizations
+that the Erlang VM provides.
+
+### Can the Apache implementation achieve similar performance?
+
+It could, but they'd have to write similar Elixir code to what we're writing,
+and do it in C++. They would also have to re-implement `quote` and `unquote`
```
jparise: Link to `quote` / `unquote` in the Elixir docs?

 - https://hexdocs.pm/elixir/Kernel.SpecialForms.html#quote/2
- https://hexdocs.pm/elixir/Kernel.SpecialForms.html#unquote/1

README.md
```diff
@@ -287,3 +287,54 @@ You can use these features to support additional languages, protocols, and serve
...
+in C++. This would be ...challenging, but it's not impossible. It's worth noting
+that while this is theoretically possible, the code would be much less maintainable,
+and the Apache implementation  isn't by any means slow. In the above benchmark,
+you can see that it serializes 1000 user objects in roughly 12ms, and deserializes
+1000 users in 21ms. If you're happy with this level of performance, then the
+Apache code should be fine.
```
dantswain: I'm not sure this particular Q/A is necessary.  It's basically a restatement of the first one but is sort of negative towards Apache/thrift/C++ and makes some subjective assertions that aren't really provable.  You could maybe take some of the points here and make them positive statements in the first A?  Like you can cite the performance speedup and point out that this library was built in Elixir using Elixir idioms and designed to be maintainable code that produces Elixir-idiomatic code (which sounds good without casting shade on C++ :P).

README.md
```diff
@@ -287,3 +287,54 @@ You can use these features to support additional languages, protocols, and serve
...
+
+
+### What tradeoffs have you made to get this performance?
```
dantswain: Has anyone actually asked about this?  Maybe it would be better to say something like "What does this library not implement?" in which case you could point out that we don't implement, say, the compact binary protocol (neither does the Erlang lib fwiw) or JSON protocol.  The Protocol/Transport/Processor breakdown always seemed really artificial and confusing to me... at the end of the day I feel like all the user cares about is if they can use the protocol(s) and server(s) they need.

README.md
```diff
@@ -287,3 +287,54 @@ You can use these features to support additional languages, protocols, and serve
...
+Processor). This means that if you need to create a new Protocol or Transport, you need to
+integrate it into this project.
+
+In practice, we've only used a combination of the Binary Protocol and Framed server but are
+more than willing to take PRs to add more. We intend to also add TMux and Finagle servers
+in the future.
+
+### What server do you use?
+
+Presently, we're using [ranch](https://github.com/ninenines/ranch) as a building block for
```
dantswain: I feel like this should be more of an assertive statement we make in the part of the documentation that describes the server, rather than something you can find if you go looking for it.

README.md
scohen: Not sure why saying that doing this in C++ would be harder than doing this in elixir is casting shade. Elixir has macros that make writing elixir in elixir easy.

README.md
scohen: I wanted to talk about why we're different from the rest of the thrift implementations. Maybe I've missed the mark.

(and to be fair, no one has asked about anything, since this isn't released ;) )

README.md
pguillory: I agree that this paragraph seems unnecessary. It's kind of confrontational and distracting. It would be preferable to keep the message positive. Here's a pure Elixir implementation of Thrift, it's fast... sells itself.

scohen: Yes, but "Why don't you integrate it into the apache project?" 
Answer: We can't, easily.

dantswain: I'd say why even address that unless it comes up?  It's fairly clear that this is a fundamentally different animal than the "official" thrift implementation.

jparise:
@dantswain Any other comments, or should @scohen move forward with merging?

dantswain:
:shipit:

PR #165, opened 2017-01-08
Ensure field names are valid elixir function names

### LINKED ISSUE #161: It's possible to generate field names that do not compile as valid elixir
Our team tends to deprecated thrift fields by prepending `DEPRECATED_` to the name -

```
struct Thing {
  1: i32 id,
  2: string DEPRECATED_phone_number
}
```
This generates a struct with an invalid field name: `defstruct(id: nil, DEPRECATED_phone_number: nil)` which causes the generated file to fail to compile.

dantswain:
This deserves automatic approval just because I got to define a function called `atomic_snake` :laughing:

Caveat: See https://github.com/elixir-lang/elixir/issues/5627 - I sort of expected `Macro.underscore/1` to normalize things but it has unexpected behavior (to me, at least).  If this does turn out to be a bug, we could simplify the function a bit.  The workaround that I used _should_ still work either way.

Closes #161

### DIFF
--- a/lib/thrift/parser/conversions.ex
+++ b/lib/thrift/parser/conversions.ex
@@ -6,6 +6,9 @@ defmodule Thrift.Parser.Conversions do
     List.to_atom(l)
   end
 
+  # convert a charlist to a snake_case atom
+  #   e.g., 'FooBar', 'foo_bar', 'fooBar', and 'FOO_BAR'
+  #   should all produce :foo_bar
   def atomic_snake(nil), do: nil
   def atomic_snake(l) when is_list(l) do
     l
--- a/lib/thrift/parser/conversions.ex
+++ b/lib/thrift/parser/conversions.ex
@@ -6,6 +6,16 @@ defmodule Thrift.Parser.Conversions do
     List.to_atom(l)
   end
 
+  def atomic_snake(nil), do: nil
+  def atomic_snake(l) when is_list(l) do
+    l
+    |> List.to_string
+    |> String.split("_")
+    |> Enum.map(&Macro.underscore/1)
+    |> Enum.join("_")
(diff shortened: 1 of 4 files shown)


dantswain:
Weird.  The build failed on my branch with Elixir 1.4 but passed here?  https://travis-ci.org/dantswain/elixir-thrift/jobs/189995590

lib/thrift/parser/conversions.ex
```diff
@@ -6,6 +6,16 @@ defmodule Thrift.Parser.Conversions do
     List.to_atom(l)
   end
 
+  def atomic_snake(nil), do: nil
```
jparise: This could use a little `@doc` and/or `@spec` to describe its behavior. It's only semi-obvious from its name. ⚛️  🐍

dantswain:
@jparise I've added a comment.  I opted for no `@doc` because this module is `@moduledoc false` (I think rightly so), and none of the other functions in this module have `@spec`s so it would be inconsistent (I could do a follow-up and add those, which should be pretty straightforward).

jparise:
@dantswain That looks good to me. Anything document-y is helpful for this case. Thanks!

dantswain:
@pguillory AFAICT think you're most familiar with the way the parser works - are there any other spots where we should be using this instead of `atomize/1`?

pguillory:
@dantswain I'm not sure actually. I see we use atomify a lot in parser/models.ex. Wouldn't be surprised if the rest should be changed as well.

dantswain:
Got pulled in other directions for the last couple days.  I'm going to merge this since it's already been reviewed and if I find other things that need to be changed over from `atomize/1` then I'll submit a new PR for those.

PR #168, opened 2017-01-09
Handled binary type in Behviour

scohen:
The behaviour module didn't properly handle the binary type.

@fire FYI

### DIFF
--- a/lib/thrift/generator/behaviour.ex
+++ b/lib/thrift/generator/behaviour.ex
@@ -17,6 +17,8 @@ defmodule Thrift.Generator.Behaviour do
     Union,
   }
 
+  require Logger
+
   def generate(schema, service) do
     file_group = schema.file_group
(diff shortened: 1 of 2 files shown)


lib/thrift/generator/behaviour.ex
```diff
@@ -107,4 +110,10 @@ defmodule Thrift.Generator.Behaviour do
       %{unquote(key_type) => unquote(val_type)}
     end
   end
+  defp typespec(unknown_typespec, _) do
+    Logger.error("Unknown type: #{inspect unknown_typespec}. Falling back to any()")
+    quote do
+      any
```
jparise: It's interesting that these differ:

```elixir
iex(1)> quote do: any
{:any, [], Elixir}
iex(2)> quote do: any()
{:any, [], []}
```

PR #180, opened 2017-01-12
Consolidate binary protocol field type constants

jparise:
We previously defined nearly identical sets of constants representing
the binary protocol's field types as attributes in each module that
required them. This duplication isn't great for long-term maintenance.

This approach defines a single set of type constants as macros in a
new `Thrift.Protocol.Binary.Type` module. Using macros lets us preserve
all of the compile-time benefits we were getting from the previous
approach while still supporting code sharing.

### DIFF
--- a/lib/thrift/generator/struct_binary_protocol.ex
+++ b/lib/thrift/generator/struct_binary_protocol.ex
@@ -49,18 +49,7 @@ defmodule Thrift.Generator.StructBinaryProtocol do
     Union,
   }
 
-  @bool 2
-  @byte 3
-  @double 4
-  @i16 6
-  @i32 8
-  @i64 10
-  @string 11
-  @struct 12
-  @union 12
-  @map 13
-  @set 14
-  @list 15
+  require Thrift.Protocol.Binary.Type, as: Type
 
   @doc """
   Generate a deserializer for a Thrift struct, union or exception.
@@ -96,10 +85,10 @@ defmodule Thrift.Generator.StructBinaryProtocol do
 
   defp field_deserializer(:bool, field, name, _file_group) do
     quote do
-      defp unquote(name)(<<unquote(@bool), unquote(field.id)::16-signed, 1, rest::binary>>, acc) do
+      defp unquote(name)(<<unquote(Type.bool), unquote(field.id)::16-signed, 1, rest::binary>>, acc) do
         unquote(name)(rest, %{acc | unquote(field.name) => true})
       end
-      defp unquote(name)(<<unquote(@bool), unquote(field.id)::16-signed, 0, rest::binary>>, acc) do
+      defp unquote(name)(<<unquote(Type.bool), unquote(field.id)::16-signed, 0, rest::binary>>, acc) do
         unquote(name)(rest, %{acc | unquote(field.name) => false})
       end
     end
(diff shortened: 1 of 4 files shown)


jparise:
I _think_ this is the best approach, but please offer alternates if you have them.

Also, I went back and forth on whether the namespace should be named `Type`, `Types`, `TType` (per the Thrift C++ source code), or `TypeID`. They all seem fine to me, so if you have a strong opinion, I'm happy to rename it.

lib/thrift/protocol/binary.ex
```diff
@@ -10,35 +10,24 @@ defmodule Thrift.Protocol.Binary do
...
+  defp type_id(:byte),      do: Type.byte
+  defp type_id(:i16),       do: Type.i16
+  defp type_id(:i32),       do: Type.i32
+  defp type_id(:i64),       do: Type.i64
+  defp type_id(:double),    do: Type.double
+  defp type_id(:string),    do: Type.string
+  defp type_id(:struct),    do: Type.struct
+  defp type_id({:map, _}),  do: Type.map
+  defp type_id({:set, _}),  do: Type.set
+  defp type_id({:list, _}), do: Type.list
```
dantswain: Could/should these and the `type_id/2` functions in struct_binary_protocol.ex be functions in the `Type` module?

jparise: I thought of doing that and wanted to get feedback on the initial approach first.

One "awkward" aspect of function in `struct_binary_protocol.ex` is that they use generator-specific types (e.g. `%Union{}`), and referencing those types from `Thrift.Protocol.` seems a little yucky.

jparise: @dantswain, I moved just the functions from `binary.ex` over to the new module. What do you think?

lib/thrift/generator/struct_binary_protocol.ex
```diff
@@ -912,23 +901,22 @@ defmodule Thrift.Generator.StructBinaryProtocol do
...
+  defp type_id(:double, _file_group),       do: Type.double
+  defp type_id(:string, _file_group),       do: Type.string
+  defp type_id(:binary, _file_group),       do: Type.string
+  defp type_id({:map, _}, _file_group),     do: Type.map
+  defp type_id({:set, _}, _file_group),     do: Type.set
+  defp type_id({:list, _}, _file_group),    do: Type.list
+  defp type_id(%TEnum{}, _file_group),      do: Type.i32
+  defp type_id(%Struct{}, _file_group),     do: Type.struct
+  defp type_id(%Exception{}, _file_group),  do: Type.struct
+  defp type_id(%Union{}, _file_group),      do: Type.struct
```
dantswain: How about the non-special cases of these get handled as a fall-through function clause?

```
defp type_id(type, _file_group), do: Type.of(type)
```

jparise: That's a great idea.

lib/thrift/protocol/binary/type.ex
```diff
@@ -0,0 +1,35 @@
...
+  def of(:bool),      do: bool()
+  def of(:byte),      do: byte()
+  def of(:i8),        do: i8()
+  def of(:i16),       do: i16()
+  def of(:i32),       do: i32()
+  def of(:i64),       do: i64()
+  def of(:double),    do: double()
+  def of(:binary),    do: string()
+  def of(:string),    do: string()
+  def of(:struct),    do: struct()
```
pguillory: Suspect that `:struct` doesn't make sense here. (Though I see you've copied it from the other file.) Something can't be of type `:struct`. It has to be a specific instance of `%Struct{}`. If anything calls `Type.of(:struct)`, it would be reasonable to let it fail to match.

jparise: I agree, and I've also removed a reference to `:union` that no longer makes sense for the same reason.

PR #182, opened 2017-01-13
Support binary default values

pguillory:


test/generator/binary_protocol_test.exs
```diff
@@ -445,6 +446,7 @@ defmodule Thrift.Generator.BinaryProtocolTest do
     5: i32 i32_val = ConstI32,
     6: i64 i64_val = ConstI64,
     7: string string_val = ConstString,
+    71: binary binary_val = ConstBinary,
```
scohen: do you check the value of that ?

Issue #57, opened 2016-11-30
Better default namespaces


scohen:
If you don't specify an elixir namespace, the generator puts files into the root namespace. This is troublesome, as different thrift modules can specify the same name for a struct / union / enum. 

If no namespace is specified, we should use the thrift namespace as a default.

dantswain:
@scohen can you clarify what you mean by "the thrift namespace"?  Do you mean, literally, "Thrift" or whatever namespace is specified in the IDL file?

pguillory:
Tried it on some of the other generators (cpp, java, php) and found that they put things in the root namespace if there is no 'namespace' annotation for the given language. I think we should stick with that behavior.

scohen:
@pguillory, but in our instance, that implementation is _broken_. If you put them in the root namespace, one version will overwrite the other.

pguillory:
What's the current thinking on this? We're adding `elixir` namespaces to our schemas?

scohen:
I'm still of the opinion that not having a namespace shouldn't break things. 

Imagine this scenario:
I have two files `service_1.thrift` and `service_2.thrift`.
They both have a struct `RequestContext`.

The last one that compiles wins. There can be only one request_context.ex. Now the service defined by `service_1.thrift` is broken. 

The thing is, all of the other thrift generators will give you a default namespace of `service_1` and `service_2`, so they won't have this problem. 

It's worth noting that this is not a hypothetical problem.

scohen:
I'm also finding myself going through our source code and adding 
```
namespace elixir Maia
```
in a file called `maia.thrift`.

jparise:
Another approach would be **requiring** the `elixir` namespace to be present in all schema files. I don't think that's the right solution, but I wanted to mention it here for completeness.

pguillory:
Here's what some of the other generators are doing when you don't specify a namespace.

- cpp: root namespace
- erl: creates a module, seems to ignore namespaces completely
- go: creates a package
- java: root namespace
- js: root namespace
- php: root namespace
- py: creates a module, i.e. file
- rb: root namespace

What we're doing at the moment is actually very similar to the Java generator, which generates a file per class. It overwrites the generated file if different thrift files use the same type name.

jparise:
We could continue with our current (Java-like) behavior but also emit a warning about using the default root namespace.

scohen:
[...]

PR #184, opened 2017-01-13
Catch name collisions within type

pguillory:
If two names of the same type (for example, two structs) exist in the same
file, one of them will be discarded silently at parse time. This causes it to
raise an exception.

See #57.

### DIFF
--- a/lib/thrift/parser/models.ex
+++ b/lib/thrift/parser/models.ex
@@ -474,13 +474,8 @@ defmodule Thrift.Parser.Models do
       case map[key] do
         nil ->
           Map.put(map, key, value)
-        existing_value ->
-          raise """
-          Name collision!
-          Name:         #{inspect key}
-          First value:  #{inspect value}
-          Second Value: #{inspect existing_value}
-          """
+        _ ->
+          raise "Name collision: #{key}"
       end
     end
   end
--- a/lib/thrift/parser/models.ex
+++ b/lib/thrift/parser/models.ex
@@ -435,31 +435,31 @@ defmodule Thrift.Parser.Models do
     end
 
     defp merge(schema, %Constant{} = const) do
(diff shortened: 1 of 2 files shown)


lib/thrift/parser/models.ex
```diff
@@ -469,6 +469,20 @@ defmodule Thrift.Parser.Models do
...
+    defp put_new_strict(map, key, value) do
+      case map[key] do
+        nil ->
+          Map.put(map, key, value)
+        existing_value ->
+          raise """
+          Name collision!
+          Name:         #{inspect key}
+          First value:  #{inspect value}
+          Second Value: #{inspect existing_value}
```
jparise: `value` vs `Value`

Also, "First" here is actually the new ("second") value to be inserted. Not sure that really matters, but it could be potentially be clearer.

dantswain: Agree "First" and "second" are potentially confusing (what is the ordering?).  Do we happen to have line information available to print out?

dantswain:
Love the idea of this.  Should it be limited to structs and in the same thrift file?  Could a struct and a const collide?  A struct and an enum?  Could they come from separate files but have the same namespace set?

scohen:
Agree with @dantswain, anything can collide within a file, not just things of the same kind.

pguillory:
Created #185 to catch other collisions.

pguillory:
I cleaned up the exception message to just "Name collision: Foo". I think this is a decent hint to the user. If they're seeing this error they probably just added a Foo, now they know there was already a Foo. Or if they're running this for the first time they'll grep for "Foo" and see multiple definitions.

pguillory:
By the way it would be preferable to include more information like the file name in which the collision occurred. It would be a more complex PR. This PR does the simple thing first, picks some low hanging fruit, and doesn't hinder us from doing a more ideal implementation later.

PR #185, opened 2017-01-14
Catch name collisions between types

pguillory:
See discussion in #57.

### DIFF
--- a/lib/thrift/parser/file_group.ex
+++ b/lib/thrift/parser/file_group.ex
@@ -47,9 +47,14 @@ defmodule Thrift.Parser.FileGroup do
 
   def add(file_group, parsed_file) do
     file_group = add_includes(file_group, parsed_file)
+    resolutions =
+      if Map.has_key?(file_group.parsed_files, parsed_file.name) do
+        file_group.resolutions
+      else
+        Resolver.add(file_group.resolutions, parsed_file)
+      end
     new_parsed_files = Map.put(file_group.parsed_files, parsed_file.name, parsed_file)
     new_schemas = Map.put(file_group.schemas, parsed_file.name, parsed_file.schema)
-    resolutions = Resolver.add(file_group.resolutions, parsed_file)
 
     %__MODULE__{file_group |
                 parsed_files: new_parsed_files,
--- a/lib/thrift/parser/resolver.ex
+++ b/lib/thrift/parser/resolver.ex
@@ -39,7 +39,22 @@ defmodule Thrift.Parser.Resolver do
     end)
 
     resolutions
-    |> Map.merge(new_type_mappings)
-    |> Map.merge(new_value_mappings)
+    |> merge_new(new_type_mappings)
+    |> merge_new(new_value_mappings)
+  end
+
+  # Similar to Map.merge but raises an error if there are any duplicate keys.
+  defp merge_new(existing_map, new_map) do
+    existing_keys = Map.keys(existing_map) |> MapSet.new
+    new_keys = Map.keys(new_map) |> MapSet.new
+
+    unless MapSet.disjoint?(existing_keys, new_keys) do
+      shared_keys = new_keys
+      |> MapSet.intersection(existing_keys)
+      |> Enum.map_join(", ", &to_string/1)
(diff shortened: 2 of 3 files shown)


lib/thrift/parser/file_group.ex
```diff
@@ -47,9 +47,12 @@ defmodule Thrift.Parser.FileGroup do
 
   def add(file_group, parsed_file) do
     file_group = add_includes(file_group, parsed_file)
+    resolutions = case Map.has_key?(file_group.parsed_files, parsed_file.name) do
```
jparise: I might use an `if` here instead because this is a `true / false` condition.

lib/thrift/parser/resolver.ex
```diff
@@ -30,6 +30,16 @@ defmodule Thrift.Parser.Resolver do
       end
     end)
 
-    Map.merge(state, new_mappings)
+    existing_keys = Map.keys(state) |> MapSet.new
+    new_keys = Map.keys(new_mappings) |> MapSet.new
+    shared_keys = MapSet.intersection(existing_keys, new_keys) |> MapSet.to_list
```
jparise: Do we need to convert this MapSet to a list? It looks like the following condition is just a emptiness test, which you can perform using `MapSet.size/1`.

lib/thrift/parser/resolver.ex
```diff
@@ -30,6 +30,16 @@ defmodule Thrift.Parser.Resolver do
...
+
+    case shared_keys do
+      [] ->
+        Map.merge(state, new_mappings)
+      [_ | _] ->
+        shared_keys_string = Enum.map_join(shared_keys, ", ", &to_string/1)
+        raise "Name collision: #{shared_keys_string}"
```
jparise: Could this include any more context, such as the name of the `.thrift` file?

lib/thrift/parser/resolver.ex
pguillory: Yeah will do. That will be more clear.

For context, I often write similar code because it's O(1) to test whether a list has zero or 1+ elements, but it's O(n) to get its exact size. But this is not performance sensitive code, and anyway it's usually zero, so let's favor clarity.

lib/thrift/parser/resolver.ex
pguillory: It's got that, sorta. For example in the test the key is "dupes.Foo" because the thrift file is "dupes.thrift".

lib/thrift/parser/resolver.ex
pguillory: Done.

pinterest-ebert:

* 6 fixed issues! 🎉

But beware that this branch is **[5 commits behind the `pinterest:thrift_tng` branch](https://github.com/pinterest/elixir-thrift/compare/pguillory:resolver_collisions...pinterest:thrift_tng)**, and a review of an up to date branch would produce more accurate results.

You can see more details about this review at https://ebertapp.io/github/pinterest/elixir-thrift/pulls/185.

lib/thrift/parser/resolver.ex
```diff
@@ -32,14 +32,13 @@ defmodule Thrift.Parser.Resolver do
...
-    shared_keys = MapSet.intersection(existing_keys, new_keys) |> MapSet.to_list
+    shared_keys = MapSet.intersection(existing_keys, new_keys)
 
-    case shared_keys do
-      [] ->
-        Map.merge(state, new_mappings)
-      [_ | _] ->
-        shared_keys_string = Enum.map_join(shared_keys, ", ", &to_string/1)
-        raise "Name collision: #{shared_keys_string}"
+    if MapSet.size(shared_keys) > 0 do
```
jparise: I think you could also use `Enum.empty?(shared_keys)` here.

pguillory: `MapSet.size` was your suggestion. It was a good suggestion. This code is fine.

pguillory:
Rebased and modified it to handle a merge conflict.

jparise:
@pguillory, would you still like to move forward with this?

PR #186, opened 2017-01-17
Reintroduce the 'thrift.generate' mix task

### LINKED ISSUE #181: Thrift compiler being run on test files
When I try to run a specific test in my project by name, the compile.thrift gets invoked on the test file. The first line of my test output is then a parse error, where it's trying to parse the Elixir source code as Thrift.
```bash
$ mix test test/whatever.exs
Failed to parse test/whatever_test.exs: Error parsing thrift file test/whatever_test.exs on line 6: {:illegal, '@'}
<test continues to run...>
```
I tried modifying the run method to raise an error
```bash
~/code/plex (master)  $ mix test test/whatever_test.exs
==> thrift
Compiling 1 file (.ex)
** (RuntimeError) boom, args = ["test/whatever_test.exs"]
    lib/mix/tasks/compile.thrift.ex:38: Mix.Tasks.Compile.Thrift.run/1
    (mix) lib/mix/task.ex:296: Mix.Task.run_task/3
    (elixir) lib/enum.ex:1184: Enum."-map/2-lists^map/1-0-"/2
    (mix) lib/mix/tasks/compile.all.ex:19: anonymous fn/1 in Mix.Tasks.Compile.All.run/1
    (mix) lib/mix/tasks/compile.all.ex:37: Mix.Tasks.Compile.All.with_logger_app/1
    (mix) lib/mix/task.ex:296: Mix.Task.run_task/3
    (mix) lib/mix/tasks/compile.ex:85: Mix.Tasks.Compile.run/1
    (mix) lib/mix/task.ex:296: Mix.Task.run_task/3
```

jparise:
My previous attempt to consolidate these two mix tasks into one was
unsuccessful due to the way mix tasks receive their command line
arguments. Specifically, we don't have a good way to differentiate
between e.g. a `mix test` command line containing test script filenames
and a `mix compile.thrift` command line naming Thrift schema files. When
the `:thrift` compiler is included, it will also run in the `mix test`
task flow and get confused about its inputs.

We now have two distinct and purpose-built mix tasks:

 - `compile.thrift` - intended for use in the Mix.compilers list
 - `thrift.generate` - intended for interactive command-line use

Fixes #181


dantswain:
I think this is OK but extra care should be taken to make sure that the documentation is clear on the existence of both tasks and their usage.

jparise:
> I think this is OK but extra care should be taken to make sure that the documentation is clear on the existence of both tasks and their usage.

Yeah, that was my original motivation to have them consolidate in one task. The things I've done to make the distinction clear:

1. Only `thrift.generate` has a `@shortdoc` attribute, so it will be the only task to show up in the `mix help` list.
2. The `@moduledoc` description for `compile.thrift` now specifically says it's a compiler task.
3. `compile.thrift` no longer does anything with command line files so it can't mistakenly be used to manually generate output files.

dantswain:
> 1. Only thrift.generate has a @shortdoc attribute, so it will be the only task to show up in the mix help list.

Is that consistent with how other compile tasks behave?  My gut feeling is it should be included for consistency-sake but I don't see any other compile tasks in `mix help`, so maybe it makes sense.

jparise:
> > Only thrift.generate has a @shortdoc attribute, so it will be the only task to show up in the mix help list.

> Is that consistent with how other compile tasks behave? My gut feeling is it should be included for consistency-sake but I don't see any other compile tasks in mix help, so maybe it makes sense.

As best as I can tell (from a survey of the stock mix compile.* tasks), yes, that is the convention.

lib/mix/tasks/thrift.generate.ex
```diff
@@ -0,0 +1,68 @@
...
+
+    * `:thrift_output` - output directory into which the generated Elixir
+      source file will be generated. Defaults to `"lib"`.
+  """
+
+  @spec run(OptionParser.argv) :: :ok
+  def run(args) do
+    {opts, files} = OptionParser.parse!(args,
+      aliases: [o: :out, v: :verbose],
+      switches: [output: :keep, verbose: :boolean])
```
pguillory: Supposed to be `out: :keep`? Why not `:string` anyway?

lib/mix/tasks/thrift.generate.ex
```diff
@@ -0,0 +1,68 @@
...
+      thrift_file
+      |> parse!
+      |> Thrift.Generator.generate!(output_dir)
+
+    if opts[:verbose] do
+      files =
+        generated_files
+        |> Enum.sort
+        |> Enum.join(" ")
+      Mix.shell.info "Wrote #{files}"
```
pguillory: What's the reasoning behind putting them all on the same line? Think it would be easier to read with one per line.

Also needs an `Enum.uniq`. The same file can be generated multiple times in case of a diamond dependency.

lib/mix/tasks/thrift.generate.ex
jparise: Woops, I'm not sure how that got so goofed up. Thanks!

lib/mix/tasks/thrift.generate.ex
jparise: I think we can evolve the output format, but I implemented multiple-line output per your suggestion because it is easier to read.

Thanks for the `Enum.uniq` suggestion. It also made me realized that, when we parse multiple input files with common includes, we end up redundantly (re)generating the common output files (once per input schema file). We should revisit that separately.

PR #189, opened 2017-01-24
Add `values/0` to generated enum module

dantswain:
This came up with one of my use cases where I needed the list of values in the enum.  Seems a reasonable counterpoint to `names/0`.


dantswain:
Good point @pguillory.  In theory this is possible with _any_ function we're defining here, not just `names` and `values`.  Would it make sense to put the macros in a submodule?  It wouldn't be as ergonomic, but you could guarantee no collisions.

I was also thinking there should be a `name_to_id` function that has a spec like `name_to_id(atom) :: integer`

scohen:
`enumeration_names` `enumeration_values` ?

jparise:
I think using a sub-module namespace for our additional functions might be the best approach. Given a enum named `Types`, we would generated `Types.Meta.names/0`, `Types.Meta.values/0`, etc.

dantswain:
For context, after having converted a project of moderate complexity from using thrash to using thrift_tng, one of the biggest changes was changing every instance of `MyEnum.id(:some_name)` to `MyEnum.some_name`. This also meant adding `require MyEnum` to every module that used it. 

I think @jparise is right - there has to be a module isolating the `.name` macros from the other functions otherwise there will always be the possibility of name collision (as unlikely as it is for someone to have a value named `id_to_name`).

Another alternative is to put the name macros in a submodule and keep the bare functions in the main generated module. I.e., if you want to use the names you have to `require MyEnum.Names`. Maybe there's a more ergonomic solution where you `use MyEnum` and that injects the name macros somehow? It could have an option to namespace them for the cases where someone has name collisions.

jparise:
> Another alternative is to put the name macros in a submodule and keep the bare functions in the main generated module. I.e., if you want to use the names you have to require MyEnum.Names. Maybe there's a more ergonomic solution where you use MyEnum and that injects the name macros somehow? It could have an option to namespace them for the cases where someone has name collisions.

I think that's a promising idea. I like how it would give more control over how the names are imported, too. Something like:

```elixir
use MyName, only: [:names, :dates]
```

pguillory:
It's just zero-arity functions that collide with the generated macros. How about `meta(:names)` and `meta(:values)`?

And actually the Thrash approach of `MyEnum.id(:some_name)` seems fine too.

I'm wary of `use` macros that set things up for you. It's expedient in the short term but adds a bit of magic for future maintainers, one more thing they have to figure out. Looking at a big, old codebase, it's nice when things are very consistently and explicitly laid out, even if it's a bit verbose and repetitive.

dantswain:
OK, I may have gone a little scope creep on this.  I liked @pguillory's suggestion of `meta` but then thought it might be helpful to also add `meta(:name_to_value_lookup)` which begs for `meta(:value_to_name_lookup)` which led to adding `name_to_value/1` and `name?/1` functions.

jparise:
I like the `meta/1` approach. It feels like a good compromise.

I'm -0 on the two additional map `_lookup`'s, but that's just because I haven't encountered the need for them myself and I generally prefer a smaller API surface. The `member?/1` function covers my primary use case of testing inclusion.

Iteration/enumeration is also useful, and you provide for this, but I'm not sure if we need both forward and inverse maps for that case (or even maps at all, instead of lists).

dantswain:
> I'm -0 on the two additional map _lookup's, but that's just because I haven't encountered the need for them myself and I generally prefer a smaller API surface. The member?/1 function covers my primary use case of testing inclusion.

I could be convinced.  I don't think I've run across a use case for it either, it's just something I was generating in Thrash.  As you say, iteration can be done via the names/values lists.

`name_to_value/1` OTOH is something I have definitely needed.  It's basically the same functionality as the macros, but "dynamic" in the sense that the name doesn't need to be known at compile time - i.e., I could have a function that accepts a `:name` and I need to grab the corresponding value.

dantswain:
Bump.  Not sure under what conditions it is copacetic for me to merge.

jparise:
I think my opinion is that we should leave out `meta(:name_to_value_map)` and `meta(:value_to_name_map)` for now. The rest of your additions look good.

I don't feel strongly so feel free to merge whichever way you decide.

lib/thrift/generator/enum_generator.ex
```diff
@@ -14,17 +14,39 @@ defmodule Thrift.Generator.EnumGenerator do
...
+      quote do
+        def name_to_value(unquote(enum_name)), do: {:ok, unquote(value)}
+      end
+    end)
+
     names = enum.values
     |> Keyword.keys
     |> Enum.map(&to_name/1)
 
+    name_to_value_map = enum.values
```
jparise: Are `name_to_value_map` and `value_to_name_map` now unused?

dantswain: Oops. Good catch.

dantswain:
Can haz merge?  (Or a confirmation that it's copacetic for me to merge given green build and approvals?)

Issue #194, opened 2017-01-25
RFC Compiler options should be grouped under a `thrift` key


dantswain:
Right now `thrift_files` and `thrift_output` have to go right in the project config's root namespace.  I think it would be cleaner if they were nested under a `thrift` key:

```
def project do
  # ..
  thrift: [
    thrift_files: Path.wildcard("thrift/*.thrift"),
    thrift_output: "gen-lib"
  ]
end
```

Doing this it might make sense to change the names of the parameters to not have a redundant "thrift_" prefix.

By way of prior art: This is the way dialyxir works, and it's fairly intuitive and unobtrusive.

jparise:
That's an interesting idea. I think there's prevent both ways (with the `erl*` and `elixir*` keys), but this could be a nice improvement.

I can look into this in a couple of weeks. It would be nice to hear any other thoughts in the interim.

dantswain:
It looks like the changes would be pretty easy to make and I'd be glad to do it if we get agreement.

ExDoc also uses this approach (`docs: []`).  Hex sort of uses this approach with the `package: []` key.

scohen:
I'm a fan, as long as we don't stutter `thrift: [files: ...]` as opposed to  `thrift: [thrift_files:...]`

jparise:
Implemented by #197

PR #197, opened 2017-01-25
Group compiler options under a :thrift key

jparise:
This change moves all compiler options under a single top-level `thrift`
configuration key that contains a Keyword list of individual options.
`thrift_files` and `thrift_output` are now just `files` and `output_path`
within this sub-list.

The goal of this change is to tidy up the top-level configuration space
and provide for further expansion of our compiler option set, per #194.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ Next, configure the compiler using a new Keyword list under the top-level
 which defines the list of Thrift files that should be compiled.
 
 By default, the generated source files will be written to the `lib` directory,
-but you can change that using the `output` option.
+but you can change that using the `output_path` option.
 
 In this example, we gather all of the `.thrift` files under the `thrift`
 directory and write our output to the `lib/thrift/` directory:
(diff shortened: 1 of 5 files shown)


pinterest-ebert:

* 6 fixed issues! 🎉

You can see more details about this review at https://ebertapp.io/github/pinterest/elixir-thrift/pulls/197.

jparise:
The Ebert failure appears to be something on their end.

README.md
```diff
@@ -68,16 +68,23 @@ It's important to add `:thrift` *before* the `:elixir` entry. The Thrift
...
+By default, the generated source files will be written to the `lib` directory,
+but you can change that using the `output` option.
+
+In this example, we gather all of the `.thrift` files under the `thrift`
+directory and write our output to the `lib/thrift/` directory:
 
 ```elixir
-thrift_files: Mix.Utils.extract_files(["thrift"], [:thrift])
+thrift: [
+  files: Mix.Utils.extract_files(["thrift"], [:thrift]),
```
dantswain: `files` seems unspecific to me... `input`/`output_dir`?  Also `Path.wildcard("thrift/*.thrift")` is a little more transparent for example purposes than `Mix.Utils.extract_files/2` IMHO.

jparise:
@dantswain Good suggestions. I address both of them with my last two commits. I'm leaning towards keeping `files` for now because I don't think that's too ambiguous for a compiler option, and it's also what the `package` configuration list uses for its inputs. But if we can think of a much better name, I'm happy to revisit.

PR #198, opened 2017-01-26
Allow enum fields to have no default value

### LINKED ISSUE #191: Enum field getting default value when none is set
Again I need to come up with an isolated test case or prove to myself that I'm incorrect, but I wanted to document this because I have to switch gears :/  Also this is on `thrift_tng`.

I.e., 
```
enum SomeEnum {
  ONE = 1,
  TWO = 2
}

struct SomeStruct {
  1: optional SomeEnum some_enum
}
```

is giving me

```
defmodule SomeStruct do
  # ...
  defstruct(some_enum: 1)
  # ...
end
```

pguillory:
Fixes #191.

### DIFF
--- a/lib/thrift/generator/struct_generator.ex
+++ b/lib/thrift/generator/struct_generator.ex
@@ -76,9 +76,6 @@ defmodule Thrift.Generator.StructGenerator do
   defp default_value(%Constant{type: type, value: value}, type, schema) do
     default_value(value, type, schema)
   end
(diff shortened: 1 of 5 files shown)


pguillory:
@scohen @jparise

pguillory:
@dantswain Ah. Yeah that was more complicated. I should be careful trying to knock things out before breakfast.

Issue #196, opened 2017-01-25
File of constants not generating any code, causes compile error


dantswain:
(On thrift_tng)

The compile error is
```
** (Enum.EmptyError) empty error
    (elixir) lib/enum.ex:1590: Enum.reduce/2
    (mix) lib/mix/utils.ex:80: Mix.Utils.stale_stream/2
    (mix) lib/mix/utils.ex:66: Mix.Utils.stale?/2
    (elixir) lib/enum.ex:787: anonymous fn/3 in Enum.filter/2
    (elixir) lib/enum.ex:1623: Enum."-reduce/3-lists^foldl/2-0-"/3
    (elixir) lib/enum.ex:787: Enum.filter/2
    lib/mix/tasks/compile.thrift.ex:42: Mix.Tasks.Compile.Thrift.run/1
    (mix) lib/mix/task.ex:296: Mix.Task.run_task/3
```

The error appears to occur in this function in the compile task when `targets` comes back as an empty list:
```
  defp stale?(%FileGroup{initial_file: thrift_file} = group, output_dir) do
    targets =
      group
      |> Thrift.Generator.targets
      |> Enum.map(&Path.join(output_dir, &1))
    Mix.Utils.stale?([thrift_file], targets)
  end
```

The file I'm parsing just contains some `const i32` definitions that aren't referenced in the other thrift files anywhere (they get used in applications that use this set of thrift files).  If I pass `--force` to the compiler, it gets past this stage but then my library fails to compile because the constants are undefined.

I'll try to come up with a test case that exercises this and if I see an easy fix I'll make a PR.

dantswain:
It actually doesn't look like const values are generating any code at all?

dantswain:
I have a POC solution for this.  The approach I took was, given a file "foo.thrift" with a non-empty set of constants, generate a module called `FooConstants` that defines a function for each constant that returns the value of the constant.  Is this a reasonable approach?  A couple alternatives I thought of:

* Generate `Foo.Constants` instead of `FooConstants` (`FooConstants` seems consistent with the way other thrift libs work and we don't nest any other modules aside from namespacing)
* Generate macros instead of functions (this could get hairy with quotes/unquotes and compilation order?)

pguillory:
It definitely should generate something for constants -- only reason it doesn't is that we hadn't gotten to it.

[...]

PR #202, opened 2017-01-26
Generate constants

dantswain:
This is re #196 

Still kind of a mess but I wanted to self-review and initiate discussion.


dantswain:
Cleaned up a bit, added some tests.  Up for review.

lib/thrift/generator/utils.ex
```diff
@@ -149,4 +159,82 @@ defmodule Thrift.Generator.Utils do
...
+    values = Enum.map(values, fn {name_charlist, value} ->
+      name = List.to_string(name_charlist) |> String.to_existing_atom
+      type = Map.fetch!(types, name)
+      {name, quote_value(value, type, schema)}
+    end)
+    quote do
+      %{unquote_splicing([{:__struct__, struct_module} | Keyword.new(defaults ++ values)])}
+    end
+  end
+  def quote_value(value, {:map, {key_type, value_type}}, schema) when is_list(value) do
```
dantswain: This is a new clause.  The constant map was coming in as a keyword list - I gotta wonder if maybe there's a parser bug where it should be giving us a map?  But I'm not familiar enough with the parser to be able to spot it...

lib/thrift/generator/utils.ex
```diff
@@ -149,4 +159,82 @@ defmodule Thrift.Generator.Utils do
     debug_optimization(expr, "done")
     expr
   end
+
+  @doc """
+  Returns a quoted value, resolving it in the schema if necessary
+
+  The values returned by this function are suitable for unquoting.
+  """
+  def quote_value(%ValueRef{} = ref, type, schema) do
```
dantswain: This function is just `StructGenerator.default_value/3` moved to Utils because I needed it to be public and usable by the constant generator.  I wasn't convinced that Utils was the best place for it to land but couldn't spot anything better - open to suggestions.

There is one new function clause that I commented on before (below in the file) that could be a bug wrt constant parsing, but I'm not sure.

lib/thrift/generator.ex
```diff
@@ -94,6 +100,20 @@ defmodule Thrift.Generator do
     end
   end
 
+  defp generate_const_modules(%Schema{constants: constants})
+  when constants == %{} do
```
jparise: Could this just match on the empty map directly?

```elixir
defp generate_const_modules(%Schema{constants: %{}}) do
```
```

Comment (dantswain) [inline lib/thrift/generator.ex]: Nope :/  `%{}` matches any map, this is the only way I know of to get an empty map through pattern matching.

Comment (jparise) [inline lib/thrift/generator.ex]: Ah, right. `%{}` only works in typespecs.

**Comment by dantswain** [comment]
Bump.

**Comment by scohen** [comment]
@dantswain This looks good to me; I was wondering if we should make them macros so they can be used in function pattern matches / guards and so they act like our enums.

**Comment by dantswain** [comment]
@scohen I tried generating macros but, as I kind of expected, I'm having trouble getting the escaping right.  Maybe you (or anyone else) have an idea?  

This is my modified constant_generator.ex:
```
defmodule Thrift.Generator.ConstantGenerator do
  @moduledoc false

  alias Thrift.Generator.Utils

  def generate(full_name, constants, schema) do
    macro_defs = Enum.map(constants, fn(constant) ->
      name = Utils.underscore(constant.name)
      value = Utils.quote_value(constant.value, constant.type, schema)
      quote do
        defmacro unquote(Macro.var(name, nil)), do: unquote(value)
      end
    end)
    quote do
      defmodule unquote(full_name) do
        unquote_splicing(macro_defs)
      end
    end
  end
end
```

It generates this:
```
defmodule(Thrift.Generator.ModelsTest.Businessy.Constants) do
  defmacro(important_cargo) do
    %{__struct__: Thrift.Generator.ModelsTest.Cargo, fib: [1, 1, 2, 3, 5], answers: MapSet.new('*'), bad_approximations: %{1 => "zero"}}
  end
  defmacro(timmy_doesnt_understand_goth) do
    %{__struct__: Thrift.Generator.ModelsTest.Goth, the_abyss: MapSet.new([]), going_nowhere: %{}, empty_like_my_soul: [1, 2, 3]}
  end
end%
```

Which causes a compile error:

```
(CompileError) test/generator/models_test.exs: invalid quoted expression: %Thrift.Generator.ModelsTest.Cargo{answers: #MapSet<'*'>, bad_approximations: %{1 => "zero"}, fib: [1, 1, 2, 3, 5]}
    expanding macro: Thrift.Generator.ModelsTest.Businessy.Constants.important_cargo/0
```

The error is due to the `#MapSet<'*'>` that gets inserted there, which isn't a legal literal value.

**Comment by dantswain** [comment]
Scratch that.  Putting in `Macro.escape(unquote(value))` worked.

@scohen @jparise @pguillory With the constants as macros, what do you think of putting them directly in the module?  I.e., instead of `things.thrift` having a constant `foo = bar` generating `defmacro foo, do: bar` in `Things.Constants`, just generate it directly in `Things`.

**Comment by jparise** [comment]
> With the constants as macros, what do you think of putting them directly in the module? I.e., instead of things.thrift having a constant foo = bar generating defmacro foo, do: bar in Things.Constants, just generate it directly in Things.

I can't think of a downside to that, and it does seem more obvious that way.

**Comment by dantswain** [comment]
So I pulled the constants back into a module defined by the thrift file's name, but it causes a problem that might need some discussion to resolve.

Consider this example:
```
// x_man.thrift
  struct XMan {
    1: string CodeName;
    2: optional string FirstName;
    3: optional string LastName
  }

  const XMan Wolverine = {"CodeName": "Wolverine", "FirstName": "Logan"}
  const XMan Cyclops = {"CodeName": "Cyclops", "FirstName": "Scott",  "LastName": "Summers"}
  const XMan Storm = {"CodeName": "Storm", "FirstName": "Ororo",  "LastName": "Monroe"}
```

I'd argue that one would _expect_ this to generate something like

```
defmodule WhateverNamespace.XMan do

dantswain:
This is spitballing a little, so I'm sure there are differences in implementation, but what I'm thinking is that the generation functions should accumulate a module => AST map which then gets rendered at the end of the chain.  That way consts can live alongside structs/enums/whatever.  We could also use that to catch collisions, though it's a different implementation than #185 I think.

```
  def generate_schema(schema) do
    List.flatten([
      generate_enum_modules(schema),
      generate_const_modules(schema),
      generate_struct_modules(schema),
      generate_union_modules(schema),
      generate_exception_modules(schema),
      generate_services(schema),
      generate_behaviours(schema)
    ])
  end
```

Should become something like this:

```
  def generate_schema(schema) do
    %{}
    |> generate_enums(schema)
    |> generate_consts(schema)
    # etc...
    |> generate_modules
  end
```

Then `generate_enums(module_to_ast, schema)` does `Map.update(module_to_ast, module_name, [enum: module_ast], fn(existing_module_ast) -> concat_ast(existing_module_ast, [enum: module_ast]) end)` where `concat_ast/2` checks for collisions and allows, e.g., consts to live in the same modules as structs / enums / whatever.  `generate_modules` then just collapses the mapping down to a list of modules' ASTs.

I'm thinking, too, that this would maybe be best as a separate PR.  This PR (#202) doesn't _break_ anything but it will fail to work for folks who have constants defined in the same files as other objects.  So if this seems like a reasonable approach, I'd propose to first merge this PR and then start work on a follow-up ASAP to implement a fix.

jparise:
@dantswain That sounds like a solid plan to me.

lib/thrift/generator/utils.ex
```diff
@@ -149,4 +159,90 @@ defmodule Thrift.Generator.Utils do
...
```
jparise: Cold you add a `@spec` to `quote_value/3` while you're here?

dantswain:
@jparise specs added.  I couldn't see any simple way to narrow down the first two arguments of `quote_value/3`.

jparise:
@dantswain Looks good. Feel free to merge whenever you're ready.

PR #204, opened 2017-01-27
Generate `@type t` for structs/unions/exceptions

dantswain:
It's fairly common to follow a defstruct with a `@type t`.  This is useful for dialyzer analysis and writing specs that convey useful information - e.g.,

```
@spec do_stuff(StructA.t, StructB.t) :: StructC.t
```

This is pretty useful even if we don't specify the types of the fields (in my experience that actually tends to be more trouble than it's worth).

I think eventually we should generate specs and docs for all generated code, but `@type t` is one of the most useful in my experience.


dantswain:
Bump.  Not sure under what conditions it is copacetic for me to merge.

PR #207, opened 2017-02-05
Annotate model structs with @enforce_keys

jparise:
This provides a simple compile-type guarantee that any required keys
have been specified when constructing these structs.


jparise:
> Are there other structs we should add this to?

Yes, we should ultimately add this to most of our structs.

PR #209, opened 2017-02-05
Handle unknown (invalid) compiler task options

### LINKED ISSUE #208: When elixir-thrift is embedded in a application, mix test.
Somehow compile.thrift gives up on extra options causing the mix test to fail when it is embedded in another application. 

Removing this from mix solves the problem, but isn't a permanent solution.

```
compilers: [:thrift | Mix.compilers],
```

Log. 

```
>mix test
==> snappyex
could not compile dependency :snappyex, "mix compile" failed. You can recompile this dependency with "mix deps.compile snappyex", update it with "mix deps.update snappyex" or clean it with "mix deps.clean snappyex"
==> ecto
** (Mix) Could not invoke task "compile.thrift": 4 errors found!
--no-deps : Unknown option
--no-archives-check : Unknown option
--no-elixir-version-check : Unknown option
--no-warnings-as-errors : Unknown option
```

jparise:
We were still using OptionParser.parse!/1 in the compiler task from back
when we also supported interactive usage. Unfortunately, that function
throws an exception when it encounters an unknown (invalid) option.

Instead, we use OptionParser.parse/1, which allows us to safely discard
invalid options.

Fixes #208

### DIFF
--- a/lib/mix/tasks/compile.thrift.ex
+++ b/lib/mix/tasks/compile.thrift.ex
@@ -25,7 +25,7 @@ defmodule Mix.Tasks.Compile.Thrift do
 
   @spec run(OptionParser.argv) :: :ok
   def run(args) do
-    {opts, _} = OptionParser.parse!(args,
+    {opts, _, _} = OptionParser.parse(args,
       switches: [force: :boolean, verbose: :boolean])
 
     config      = Keyword.get(Mix.Project.config, :thrift, [])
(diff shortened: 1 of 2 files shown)


jparise:
> Does this need to get merged into master? It wasn't clear if #208 was for master or thrift-tng.

The code on **master** was still using `OptionParser.parse/1` so it should be fine.

PR #211, opened 2017-02-06
Add typespecs to various utility functions

jparise:


lib/thrift/generator/utils.ex
```diff
@@ -59,9 +61,12 @@ defmodule Thrift.Generator.Utils do
     end)
   end
 
+  @spec underscore(atom) :: atom
   def underscore(a) when is_atom(a) do
     a |> Atom.to_string |> underscore |> String.to_atom
   end
+
+  @spec underscore(bitstring) :: bitstring
```
dantswain: I think this is always a binary?

jparise:
> `optimize_iolist(list) :: any` could be `optimize(iodata) :: iodata`

I don't think that's true, unfortunately. It often receives a `list` of atoms as an input, for example. But I believe we can make it return `iodata`.

dantswain:
@jparise If you merge this I'll merge it into #202 before merging to make sure I don't accidentally break specs somewhere.

PR #213, opened 2017-02-07
Update some ProtocolHandler @spec's

jparise:
start_link/5's @spec wasn't updated when it signature was last changed.
Also add a @spec to init/6 and mark some internal functions as private.

### DIFF
--- a/lib/thrift/binary/framed/protocol_handler.ex
+++ b/lib/thrift/binary/framed/protocol_handler.ex
@@ -11,12 +11,13 @@ defmodule Thrift.Binary.Framed.ProtocolHandler do
   }
   require Logger
 
-  @spec start_link(reference, pid, module, {module, module}) :: GenServer.on_start
+  @spec start_link(reference, pid, module, {module, module, Keyword.t}) :: GenServer.on_start
   def start_link(ref, socket, transport, {server_module, handler_module, tcp_opts}) do
     pid = spawn_link(__MODULE__, :init, [ref, socket, transport, server_module, handler_module, tcp_opts])
     {:ok, pid}
   end
 
+  @spec init(reference, pid, module, module, module, Keyword.t) :: :ok | no_return
   def init(ref, socket, transport, server_module, handler_module, tcp_opts) do
     :ok = :ranch.accept_ack(ref)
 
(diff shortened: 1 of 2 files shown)


lib/thrift/binary/framed/protocol_handler.ex
```diff
@@ -11,12 +11,13 @@ defmodule Thrift.Binary.Framed.ProtocolHandler do
   }
   require Logger
 
-  @spec start_link(reference, pid, module, {module, module}) :: GenServer.on_start
+  @spec start_link(reference, pid, module, {module, module, Keyword.t}) :: GenServer.on_start
```
dantswain: Is it worth defining a type for tcp_opts?  Can be helpful for documentation purposes.

dantswain: Should we tighten down what `transport` is allowed to be?  Not just in the spec but also in `start_link` - `when transport in [OptionA, OptionB]`?

jparise: Have a look at what I just pushed. I think it's a reasonable balance between the semi-generic nature of the ProtocolHandler and the fact that it's implementation is `:ranch_tcp`-specific right now.

lib/thrift/binary/framed/protocol_handler.ex
```diff
@@ -5,20 +5,26 @@ defmodule Thrift.Binary.Framed.ProtocolHandler do
...
-    pid = spawn_link(__MODULE__, :init, [ref, socket, transport, server_module, handler_module, tcp_opts])
+  @spec start_link(reference, port, transport, {module, module, transport_opts}) :: GenServer.on_start
+  def start_link(ref, socket, transport, {server_module, handler_module, transport_opts}) do
+    pid = spawn_link(__MODULE__, :init, [ref, socket, transport, server_module, handler_module, transport_opts])
     {:ok, pid}
   end
 
-  @spec init(reference, pid, module, module, module, Keyword.t) :: :ok | no_return
-  def init(ref, socket, transport, server_module, handler_module, tcp_opts) do
+  @spec init(reference, port, :ranch_tcp, module, module, :ranch_tcp.opts) :: :ok | no_return
```
dantswain: What happens when this is `:ranch_ssl`?

jparise: `init/6` doesn't support `:ranch_ssl`; I added the `:ranch_tcp` pattern match the function signature to lock that in.

We should need to write a new `:ranch_ssl`-specific implementation to support it.

I guess, given that, I should drop `:ranch_ssl` from the `transport` type for now?

dantswain: Ah I didn't realize that.

> I guess, given that, I should drop :ranch_ssl from the transport type for now?

Yeah that makes sense to me.  Maybe an issue so someone remembers to go back and support it?

jparise: Filed as #214

PR #217, opened 2017-02-08
Fix the "consistency" warnings flagged by Credo

jparise:
Also, only lint our test/support/ code and not all of our test scripts
for now.


jparise:
> Personally I prefer just writing out the aliases - makes code search much easier.

I haven't decided which one I personally prefer, but because we were already using the multi format in so many places, I think that establishes sufficient precedent for now.

PR #219, opened 2017-02-08
Support additional include paths

### LINKED ISSUE #188: Support a list of "include" search paths
The Apache Thrift compiler lets you specify a list of paths to search for included files:

```
  -I dir      Add a directory to the list of directories
                searched for include directives
```

We should add a similar capability to our generator, probably configured via a `thrift_includes: []` value.

jparise:
We previously only supported including other Thrift files whose
pathnames were relative to the current file. This change allows a list
of additional include (search) paths to be specified as part of the
project configuration (`thrift.include_paths`) or via a command line
option (`--include dir`).

Closes #188

### DIFF
--- a/lib/mix/tasks/thrift.generate.ex
+++ b/lib/mix/tasks/thrift.generate.ex
@@ -6,19 +6,24 @@ defmodule Mix.Tasks.Thrift.Generate do
   @moduledoc """
   Generate Elixir source files from Thrift schema files (`.thrift`).
 
-  A list of files can be given after the task name in order to select the
-  specific Thrift schema files to parse:
+  A list of files should be given after the task name in order to select
+  the specific Thrift schema files to parse:
 
       mix thrift.generate file1.thrift file2.thrift
 
   ## Command line options
 
-    * `-o` `--out` - set the output directory, overriding the `:output_path`
-      configuration value
-    * `-v` `--verbose` - enable verbose task logging
+    * `-I dir` / `--include dir` - add a directory to the list of
+      directory paths in which to search for included files, overriding
+      the `:include_paths` configuration value
+    * `-o dir` / `--out dir` - set the output directory, overriding the
+      `:output_path` configuration value
+    * `-v` / `--verbose` - enable verbose task logging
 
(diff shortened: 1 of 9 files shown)


lib/mix/tasks/thrift.generate.ex
```diff
@@ -6,53 +6,61 @@ defmodule Mix.Tasks.Thrift.Generate do
...
+  the specific Thrift schema files to parse:
 
       mix thrift.generate file1.thrift file2.thrift
 
   ## Command line options
 
-    * `-o` `--out` - set the output directory, overriding the `:output_path`
-      configuration value
-    * `-v` `--verbose` - enable verbose task logging
+    * `-I dir` / `--include dir` - add a directory to the list of
```
dantswain: How does one specify more than one directory?  Is it like `-I foo -I bar` or `-I foo,bar`?  Can the description make that more obvious?

jparise: It's the former, which fortunately matches the way that the `thrift` compiler works. I'll expand the description to make it clear that this option can be repeated.

lib/thrift/parser/file_group.ex
```diff
@@ -57,19 +62,29 @@ defmodule Thrift.Parser.FileGroup do
...
-    Enum.reduce(schema.includes, group, fn(include, file_group) ->
-      parsed_file = file_ref.path
-      |> Path.dirname
-      |> Path.join(include.path)
-      |> FileRef.new
-      |> ParsedFile.new
-      add(file_group, parsed_file)
+    Enum.reduce(schema.includes, group, fn(include, group) ->
+      parsed_file =
+        include.path
```
dantswain: Unrelated to this PR - has this become the preferred style?

jparise: Indenting a pipeline below the assignment/match? Yes, I started noticing this is how `vim-elixir` wanted to format my code, and then I browsed around the Elixir standard library and found the same pattern. I've grown to find it pretty readable, too.

PR #222, opened 2017-02-09
Remove the numbers.thrift test file

jparise:
The new ThriftTest.thrift includes constants, so numbers.thrift is now
redundant.


dantswain:
@jparise I added that because it explicitly tested the issue that came up in #196 where a file containing only constants would cause an error on compile.  It turns out that the reason was because no code was being generated for constants, so technically we do now test the underlying issue.  I don't feel that strongly about it either way I guess, just wanted to explain why I added it.

jparise:
Ah, that makes more sense. Let's leave it for now then. I was only removing it because I thought it was entirely redundant.

Issue #220, opened 2017-02-09
Constants and structs/etc collide and overwrite one another


dantswain:
See discussion in #202

dantswain:
Closed by #224

PR #224, opened 2017-02-09
Handle name collisions of different types, allow constants to live beside structs/etc

dantswain:
Also fix a bug in Utils.underscore (it's fixed upstream in the Elixir source but we probably want to support some of the existing Elixir versions).

This is re: #220 and could potentially also resolve #185 - I'm not sure how to test the assertion on name collisions within our test framework, though.

### DIFF
--- a/lib/thrift/generator.ex
+++ b/lib/thrift/generator.ex
@@ -93,7 +93,7 @@ defmodule Thrift.Generator do
       filename
     end)
   end
-  
+
   defp resolve_name_collisions(generated_modules) do
     Enum.reduce(generated_modules, [], fn({name, quoted}, acc) ->
       Keyword.update(
--- a/lib/thrift/generator.ex
+++ b/lib/thrift/generator.ex
@@ -81,6 +81,7 @@ defmodule Thrift.Generator do
 
   defp write_schema_to_file(generated_modules, output_dir) do
     generated_modules
+    |> resolve_name_collisions
     |> Enum.map(fn {name, quoted} ->
       filename = target_path(name)
       source = Macro.to_string(quoted)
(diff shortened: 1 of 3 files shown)


lib/thrift/generator/utils.ex
```diff
@@ -77,8 +77,14 @@ defmodule Thrift.Generator.Utils do
     a |> Atom.to_string |> underscore |> String.to_atom
   end
 
+  # NOTE this properly handles SCREAMING_SNAKE_CASE strings
```
jparise: Perhaps add a note that this was fixed upstream.

lib/thrift/generator.ex
```diff
@@ -93,6 +94,48 @@ defmodule Thrift.Generator do
     end)
   end
 
+  defp resolve_name_collisions(generated_modules) do
+    Enum.reduce(generated_modules, [], fn({name, quoted}, acc) ->
+      Keyword.update(
+        acc,
+        name,
+        quoted,
+        fn(existing) -> resolve_name_collision(name, existing, quoted) end
```
jparise: This could also be (shorter):

```elixir
&resolve_name_collisions(name, &1, quoted)
```

lib/thrift/generator.ex
```diff
@@ -93,6 +94,48 @@ defmodule Thrift.Generator do
...
+    end)
+  end
+
+  defp resolve_name_collision(name, q1, q2) do
+    {meta1, ast1} = get_meta_and_ast(q1)
+    {meta2, ast2} = get_meta_and_ast(q2)
+
+    context1 = Keyword.get(meta1, :context)
+    context2 = Keyword.get(meta2, :context)
+    combined_ast = [name, [do: {:__block__, [], ast1 ++ ast2}]]
```
jparise: Is this used?

lib/thrift/generator.ex
```diff
@@ -93,6 +94,48 @@ defmodule Thrift.Generator do
...
+
+    # the context will be the generating module
+    # we can combine constants into other modules, but no other combinations
+    # furthermore, we want to keep whichever context is _not_ the constant
+    # generator so that subsequent combinations will work properly
+    cond do
```
jparise: I'm unclear on why we're guaranteed that one of these is a `ConstantGenerator`.

dantswain: The generator tells us what kind of modules we're asking to combine.  We could theoretically combine any two modules but I could only think of Constant combining with other module types.  All of the other ones would try to do multiple `defstruct`s in the same module.

Except now that I think of it, Enum could probably also merge with other types?

jparise: Ah, I understand now: it's the `:defmodule` AST match that we rely on to identify this case. That makes sense, but I don't think that was obvious based on how I was reading the code.

:+1: to the TEnum case, too.

dantswain: I'll definitely update the comments to hopefully make it clearer.

I was just thinking about the TEnum case and I'm not sure it will be easy to work that out cleanly... It should work MOST of the time, but suppose someone has

```
// user_status.thrift
enum UserStatus {
  ACTIVE,
  BANNED
}

const i32 One = 1
```
It's a little convoluted, but we would end up with both a `def one` and a `defmacro one` in the same `UserStatus` module.  Resolving _that_ would be a lot more involved.  We could either just disallow merging TEnum and Constant or allow it with the caveat that this edge case could come up (why would someone define an enum and a const with the same name?).

PR #228, opened 2017-02-10
Allow the namespace option to be a string or module

dantswain:
Also normalize it before use - someone could put in "foo" and we need "Foo" so I put in a `Macro.camelize`.

This came up for me when I went to use the new option and it wasn't clear if `namespace` should be a string or a module.  This makes it so either is OK.  I could be convinced that we should be more strict and require a string, but if we do that then I think we should a) make the help text/docs clear on that, and b) raise an error saying that the input is invalid if it's either not a string or if it's not a CamelCase string.  If that's preferable, I could pivot this PR to do that pretty easily I think.

### DIFF
--- a/test/thrift/parser_test.exs
+++ b/test/thrift/parser_test.exs
@@ -1,8 +1,10 @@
 defmodule ParserTest do
-
   use ExUnit.Case, async: true
 
-  import Thrift.Parser, only: [parse: 1, parse: 2]
+  @project_root Path.expand("../..", __DIR__)
+  @test_file_dir Path.join([@project_root, "tmp", "parser_error"])
+
+  import Thrift.Parser, only: [parse: 1, parse: 2, parse_file: 2]
 
   alias Thrift.Parser.Models.Constant
   alias Thrift.Parser.Models.Exception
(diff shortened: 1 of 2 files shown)


dantswain:
Hrm.  Two of the ebert issues are redeclaring variables.  I quite often write exunit tests like this:

```
# do some shared setup

input = "some input"
expected = "some output"
assert expected == do_something(input, shared_setup_stuff)

input = "some other input"
expected = "some output"
assert expected == do_something(input, shared_setup_stuff)


input = "yet a different input"
expected = "some output"
assert expected == do_something(input, shared_setup_stuff)
```

I prefer that over creating multiple test cases because then there's only one place to have to maintain the setup (and ExUnit doesn't have a scoped setup mechanism like Rspec's `context` - arguably rightly so because Elixir isn't Ruby).  Reusing the variable names makes it easy to copy/paste and create new test cases.

I can change it to whatever style you guys prefer, I just wanted to explain my initial choice.

test/thrift/parser_test.exs
```diff
@@ -1,8 +1,10 @@
 defmodule ParserTest do
-
   use ExUnit.Case, async: true
```
jparise: I don't think this can be async anymore if we share the test file directory across individual tests.

dantswain: Woops.  I copy/pasta'd the directory name.  Fixed that and changed the `setup` block to be `setup_all`.  That should work as long as someone doesn't reuse the same file name in multiple tests, right?

jparise: Yes, I think that's correct.

PR #233, opened 2017-02-11
Serializer / Deserializer do not respect signedness

### LINKED ISSUE #206: elixir-thrift's integers are assumed to be unsigned and breaks when negative
thrift's i64 decoding may be broken
 
I get 18446744011573782016 instead of -62135769600

Example code:
```
<micmus> <<int::integer-64-signed>> = :binary.encode_unsigned(18446744011573782016)
<micmus> int => -62135769600
<iFire> iex(44)> DateTime.from_unix(int)
<iFire> {:ok, #<DateTime(0000-12-30T00:00:00Z Etc/UTC)>}
<iFire> \o/
```

This means 16, 32 and 64 signed integers are also broken according to a review of `struct_binary_protocol.ex line 132`.

scohen:
We were using size(x) rather than x-signed; apparently, size(x) is for
unsigned numbers.

Fixed, and added unit tests for boundary conditions. Fixes #206

### DIFF
--- a/lib/thrift/generator/struct_binary_protocol.ex
+++ b/lib/thrift/generator/struct_binary_protocol.ex
@@ -112,14 +112,14 @@ defmodule Thrift.Generator.StructBinaryProtocol do
   end
   defp field_deserializer(:i16, field, name, _file_group) do
     quote do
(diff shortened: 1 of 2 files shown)


jparise:
> We should consider property testing. That might have caught this.

Yes, that's what I had started in my branch to address this issue (using [excheck](https://github.com/parroty/excheck)). I think it's the best approach to these kinds of tests.

Issue #138, opened 2017-01-03
Recompile generated code after elixir-thrift dep.update


pguillory:
The current thrift.compile logic only recompiles output files based on their last_modified timestamp relative to their respective input files. However they may also need to be regenerated if elixir-thrift itself has changed. For example the client and server have generated user modules interoperating with static modules like Thrift.Client.BinaryFramed. If the latter changes, we need to make sure the former changes as well.

jparise:
I would like to add support for having the compiler task write a "manifest" that would track its generated files. That would gives us a mechanism to clean and regenerate output files. This would also be useful in the event that our code generator changes the names of the output files that it generates.

Separately, we could consider some sort of "version" that ties the generated files to a specific revision of the code generator. Some possible options:

1. **Use an incremental code version number.** This would be the most specific way to flag codegen changes, but it's also error prone because a human needs to bump this value manually.
2. **Use the package version.** This is nice because it is tied to the package release process, but it doesn't handle intra-release changes. We might be able to use a git revision instead.
3. **Use an internal BEAM version.** Perhaps in combination with the above, we could compare [BEAM module versions](http://erlang.org/doc/man/beam_lib.html#version-1) to detect changes to the code generator. This is probably too magical for use in practice and feels like it could produce both false-positives _and_ false-negatives.

pguillory:
Where would the manifest file reside, the project root? How would it handle manual compile.thrift invocations alongside automatic ones? Like would it append to the manifest? How do you clean out that stuff when you no longer use a particular .thrift file?

We could store metadata in generated files as comments at the top of the file.

```elixir
# generated_at: 2017-01-1 12:34:56
# thrift_version: abc123
# etc...
```

Determining the thrift version is still an interesting problem. I like the idea of using the beam version. There are several modules under Elixir.Thrift.* though, would we need to check them all?

What if we look at the modification times of our beam files, rather than the version? Effectively:

```bash
find . -name "Elixir.Thrift.*.beam" -exec stat -f %m {} \; | sort | tail -n1
```

And when we do the stale check, take the most recent modification from those + the .thrift file in question. Kind of hacky though. But it would be nice to have something that works for us during development and also for end users, without having to remember to bump version numbers.

jparise:
> Where would the manifest file reside, the project root?

[...]

PR #234, opened 2017-02-11
Maintain a manifest of generated files

jparise:
This is signifiant rewrite of the compiler task to support manifests,
making us a better mix citizen and adding support for `mix clean`-based
file cleanup.

Our manifest is written as a binary-encoded Erlang term to give us the
most forward-compatible file format. At the moment, we only store a
manifest version number and the list of generated files, but we may
expand this in the future to include additional dependency information
such as the "version" of the code that was used to generate these files.
See #138 for a more detailed discussion.

### DIFF
--- a/lib/mix/tasks/compile.thrift.ex
+++ b/lib/mix/tasks/compile.thrift.ex
@@ -118,11 +118,13 @@ defmodule Mix.Tasks.Compile.Thrift do
       File.mkdir_p!(output_path)
       Enum.each(removed, &File.rm/1)
 
-      Mix.Utils.compiling_n(length(stale), :thrift)
-      Enum.each(stale, fn {group, _targets} ->
-        Thrift.Generator.generate!(group, output_path)
-        verbose && Mix.shell.info "Compiled #{group.initial_file}"
-      end)
+      unless Enum.empty?(stale) do
+        Mix.Utils.compiling_n(length(stale), :thrift)
+        Enum.each(stale, fn {group, _targets} ->
+          Thrift.Generator.generate!(group, output_path)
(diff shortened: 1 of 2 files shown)


lib/mix/tasks/compile.thrift.ex
```diff
@@ -27,32 +31,40 @@ defmodule Mix.Tasks.Compile.Thrift do
...
+
+  @doc "Cleans up generated files."
+  @spec clean :: :ok | {:error, File.posix}
+  def clean, do: clean(manifest())
+  defp clean(manifest) do
+    Enum.each(read_manifest(manifest), &File.rm/1)
+    File.rm(manifest)
   end
 
+  @spec parse(Path.t, OptionParser.parsed) :: FileGroup.t
```
dantswain: Why add `@spec`s to private functions?  It doesn't actively _hurt_ anything but it is a maintenance burden... just curious.

jparise: Ironically, I added these when I played with moving some logic into another module, where these functions were public. I continued to find them useful as a way to document the internal arguments.

jparise:
Yeah, I spent a lot of time reading through mix to best understand how tasks like this should work.

- `mix clean` calls all available `clean` functions on the modules listed in `Mix.compilers`.
- `mix compile` calls all available `manifests` functions on all `compile.*` tasks.

Also, I added `@recursive true` to our compile task, which makes it run recursively when called as part of an umbrella project.

I ultimately decided not to introduce another module because the manifest bits are specific to this task and don't need to be part of our `Thrift.Generator` code. And reading the updated task end-to-end isn't too complicated, even with these changes.

PR #238, opened 2017-02-14
Include our package version in the manifest header

jparise:
This gives us additional control over when to consider existing
artifacts as "stale" and in need of regeneration. At the moment, we
require a strict version string match, but as we mature in accordance
with semver guidelines, we can switch to a more fine-grained approach
based on Version.compare/2.

We don't need to bump the manifest version in this case because ...

1. Reads against existing manifests will safely fail and result in
   desirable rebuild behavior.
2. The manifest stuff is brand new, and it's unlike many folks are
   using it yet.

### DIFF
--- a/lib/mix/tasks/compile.thrift.ex
+++ b/lib/mix/tasks/compile.thrift.ex
@@ -133,21 +133,29 @@ defmodule Mix.Tasks.Compile.Thrift do
     end
   end
 
+  defp thrift_vsn do
+    Keyword.get(Mix.Project.config, :version)
+  end
+
   @spec read_manifest(Path.t) :: [Path.t]
   defp read_manifest(manifest) do
+    header = {@manifest_vsn, thrift_vsn()}
     try do
       manifest |> File.read! |> :erlang.binary_to_term
     rescue
       _ -> []
     else
-      [@manifest_vsn | paths] -> paths
+      [^header | paths] -> paths
       _ -> []
     end
   end
 
   @spec write_manifest(Path.t, [Path.t], :calendar.datetime) :: :ok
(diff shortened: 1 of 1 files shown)


lib/mix/tasks/compile.thrift.ex
```diff
@@ -133,21 +133,29 @@ defmodule Mix.Tasks.Compile.Thrift do
     end
   end
 
+  defp thrift_vsn do
```
dantswain: This name makes me think it's the actual name of the version of thrift supported (e.g., 0.9.3), which kind of opens up a can of worms... so far we seem to be 0.9.3 compatible and thrift seems to be backwards compatible, but does anyone know if they have made any statements about potential for backwards incompatibility?  If thrift 1.0.0 was not backwards compatible with 0.9.x, how would we handle that?

jparise: Would it be clearer if I renamed this to `package_vsn`?

dantswain: Ooh I like that.  It's accurate without being overly specific.

dantswain:
@jparise I just came across this today.  We're using `thrift_tng` on a new project and have a couple devs working on it.  When I updated today to pull in the TEnum `meta` function, it caused some ripple effects where people had to force a recompile of generated code (e.g., `rm lib/generated/*`).

So long story short, I was thinking it might be useful to actually tag the manifest with the commit SHA somehow?

jparise:
Yes, this is essentially #138. I'm not sure how much work we want to put into this system to support our current pre-release development environment, but I'm sure we can find some sort of compromise. Let's discuss more on that issue.

PR #241, opened 2017-02-15
Revise our Dialyzer configuration

jparise:
This upgrades dialyxir, simplifies the set of included apps, and adds an
"ignore" rule for the leex-related warning that we otherwise can't
control.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -41,7 +41,10 @@ defmodule Thrift.Mixfile do
      package: package(),
 
      # Dialyzer
-     dialyzer: [plt_add_deps: :transitive, plt_add_apps: [:ex_unit, :mix, :public_key, :ssl]],
+     dialyzer: [
+       plt_add_deps: :transitive,
+       plt_add_apps: [:mix],
+       ignore_warnings: ".dialyzerignore"],
 
      # Docs
      name: "Thrift",
--- a/.dialyzerignore
+++ b/.dialyzerignore
@@ -1,0 +1,1 @@
+Function yyrev/2 will never be called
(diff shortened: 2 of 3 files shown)


mix.exs
```diff
@@ -41,7 +41,10 @@ defmodule Thrift.Mixfile do
      package: package(),
 
      # Dialyzer
-     dialyzer: [plt_add_deps: :transitive, plt_add_apps: [:ex_unit, :mix, :public_key, :ssl]],
+     dialyzer: [
+       plt_add_deps: :transitive,
+       plt_add_apps: [:mix],
+       ignore_warnings: ".dialyzer.warnings"],
```
dantswain: Nice, I didn't know this was an option!

mix.exs
```diff
@@ -41,7 +41,10 @@ defmodule Thrift.Mixfile do
      package: package(),
 
      # Dialyzer
-     dialyzer: [plt_add_deps: :transitive, plt_add_apps: [:ex_unit, :mix, :public_key, :ssl]],
+     dialyzer: [
+       plt_add_deps: :transitive,
+       plt_add_apps: [:mix],
```
dantswain: IIRC I had to add `:ex_unit`, `:public_key`, and `:ssl` because they weren't being pulled into the PLT.  Did that get fixed some other way?

mix.exs
dantswain: If I'm being super nitpicky, `.dialyzer.warnings` doesn't convey "these are warnings that will be ignored".  What about something like`.dialyzerignore` (in the vein of `.gitignore`, etc)?

jparise: Yes, this is a recent addition to dialyxir. I renamed the file per your suggestion.

mix.exs
jparise: You had added `:ex_unit`, and then I added `:public_key` and `:ssl` when I was adding specs for some of the ranch types. Those three no longer appear to be explicitly needed now, perhaps due to this dialyzer upgrade. I went through a few clean builds to be sure.

`:mix` is still necessary.

PR #242, opened 2017-02-16
Implement Exception.message/1 for exceptions

### LINKED ISSUE #239: Elixir-compatible exception structs
Elixir requires exception structs to define two function in compliance with the Exception behaviour: `exception/1` and `message/1`.

We generate Elixir exception structs via `defexception` for Thrift exception types. That macro always provides an `exception/1` implementation, but `message/1` is only generated if the exception is defined with a `:message` field. When it doesn't, the generated code produces warnings like this:

```
warning: undefined behaviour function message/1 (for behaviour Exception)
```

> message/1 - receives the exception struct and must return its message. Most commonly exceptions have a message field which by default is accessed by this function. However, if an exception does not have a message field, this function must be explicitly implemented.

We can accommodate this requirement in one of two ways:

1. We can insert a `:message` key into our exception structs whenever one doesn't already exist and give it a reasonable default value.
2. We can generate our own `message/1` functions for exceptions, conditionally (based on the existence of a `:message` field) or always.

(1) is the easiest thing to do in terms of code changes, but it has the disadvantage of introduce a struct field that doesn't exist in the underlying Thrift IDL.

(2) doesn't suffer from any obvious disadvantages, although it does appear to involve more complicated code changes based on how we generate the quoted module AST (unless I'm missing a more obvious way to conditionialize it).

jparise:
We generate Elixir exception structs via [`defexception`](https://hexdocs.pm/elixir/Kernel.html#defexception/1) for Thrift
exception types. That macro always provides an `exception/1`
implementation, but `message/1` is only generated if the exception is
defined with a `:message` field. When it doesn't, the generated code
produces a warning because the full [`Exception`](https://hexdocs.pm/elixir/Exception.html) behaviour hasn't been
implemented.

This change generates a `message/1` implementation for all of our Thrift
exception structs which don't have a natural `:message` field. It returns
the [`Kernel.inspect/2`](https://hexdocs.pm/elixir/Kernel.html#inspect/2) string representation of the exception, which is
both simple and unambiguous.

Fixes #239

### DIFF
--- a/test/thrift/generator/binary_protocol_test.exs
+++ b/test/thrift/generator/binary_protocol_test.exs
@@ -327,6 +327,11 @@ defmodule Thrift.Generator.BinaryProtocolTest do
     assert_serializes %Exception{val_list: [%Ex{num: 91}]},             <<15, 0, 4, 12, 0, 0, 0, 1, 3, 0, 99, 91, 0, 0>>
   end
 
+  thrift_test "exceptions provide message/1" do
+    assert Ex.message(%Ex{message: "text", num: 1}) ==
+      ~s(%Thrift.Generator.BinaryProtocolTest.Ex{message: "text", num: 1})
+  end
+
   @thrift_file name: "composite.thrift", contents: """
   struct Composite {
     1: optional map<map<byte, byte>, map<byte, byte>> map_of_maps;
--- a/test/thrift/generator/binary_protocol_test.exs
+++ b/test/thrift/generator/binary_protocol_test.exs
@@ -306,6 +306,9 @@ defmodule Thrift.Generator.BinaryProtocolTest do
      1: optional string message,
     99: optional byte num;
   }
+  exception Ex2 {
+    1: optional i16 error_code;
+  }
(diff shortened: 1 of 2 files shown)


test/thrift/generator/binary_protocol_test.exs
```diff
@@ -327,6 +327,11 @@ defmodule Thrift.Generator.BinaryProtocolTest do
     assert_serializes %Exception{val_list: [%Ex{num: 91}]},             <<15, 0, 4, 12, 0, 0, 0, 1, 3, 0, 99, 91, 0, 0>>
   end
 
+  thrift_test "exceptions provide message/1" do
+    assert Ex.message(%Ex{message: "text", num: 1}) ==
```
scohen: The context isn't clear; but doesn't the `Ex` exception already have a message defined?

jparise: It does, but per my commit summary, I've chosen to always override it with the `inspect/2`-based function.

PR #243, opened 2017-02-26
Simplify the lexer's rules

jparise:
There are multiple of small changes here that result in a large
simplification of the lexer's rules:

1. Many rules that were mapping literals to atoms have been collapsed in
   favor of generic rules that call `list_to_atom/1` on `TokenChars`.
2. Quoted string processing now uses a simpler regular expression paired
   with a processing function, giving us support for character escape
   sequences.
3. The `st_ident` token type has been removed because Smalltalk-specific
   IDL support was been deprecated a long time ago by Apache Thrift, and
   we don't generate Smalltalk namespaces in our library.

Also, the lexer unit test now verifies that we can tokenize a complete
document.


jparise:
@scohen Thanks. The only regex I changed was the one for strings. It's individually simpler than what was there before, but I found it to be more maintainable to compile the two patterns (for `'` and `"`) onto the same line.

PR #244, opened 2017-02-27
Connections couldn't be closed

scohen:
Previously, connections would reconnect if they were closed by calling
`Client.close`. This is bad behavior, and could easily lead to running
out of sockets. `Client.close` now closes the TCP connection and stops
the process.

### DIFF
--- a/test/thrift/generator/service_test.exs
+++ b/test/thrift/generator/service_test.exs
@@ -316,7 +316,7 @@ defmodule Thrift.Generator.ServiceTest do
 
   thrift_test "clients can be closed", ctx do
     ref = Process.monitor(ctx.client)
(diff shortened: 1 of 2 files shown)


test/thrift/generator/service_test.exs
```diff
@@ -315,7 +315,11 @@ defmodule Thrift.Generator.ServiceTest do
   # connection tests
 
   thrift_test "clients can be closed", ctx do
-    :ok = Client.close(ctx.client)
+    ref = Process.monitor(ctx.client)
+    Client.close(ctx.client)
```
jparise: Any reason not to continue checking the `Client.close` result?

PR #245, opened 2017-02-27
Improve the grammar's Thrift IDL compliance

jparise:
This is a near complete rewrite of our yecc grammar. It includes the
following changes:

1. Our symbols now follow Erlang's TitleCase naming convention. This
   makes them easier to differentiate from literals and tokens.
2. We no longer support set constants because they aren't supported by
   the Apache Thrift compiler.
3. We no longer support parens around a service definition's "extends"
   clause. That was introduced by mistake and isn't part of the official
   Thrift IDL specification.
4. We now support both `,` and `;` as list separators. We previously
   only supported `,` while `;` was discarded by the tokenizer.


jparise:
I think coverage decreased because I removed some now-unused test cases. I'll work on getting parser coverage back up.

PR #246, opened 2017-02-28
Fix up a number of model-related dialyzer warnings

jparise:
1. Always refer to models types using their `t` type so we get full
   dialyzer visibility into argument and return values.
2. Many field names are atoms now and shouldn't use String.t.
3. Fix the field list and enum list types, which are lists and not maps.
4. Suppress dialyzer warnings in generator/service.ex due to a mismatch
   between response field types and function return types.

### DIFF
--- a/lib/thrift/generator/service.ex
+++ b/lib/thrift/generator/service.ex
@@ -12,6 +12,12 @@ defmodule Thrift.Generator.Service do
     Struct
   }
 
+  # The response struct uses a %Field{} to represent the service function's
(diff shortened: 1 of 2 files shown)


jparise:
@scohen Yes, but it's a little bit involved because we'd probably want to precompile PLTs to avoid slowing down our builds tremendously.

Here's a good summary: http://blog.danielberkompas.com/elixir/2015/04/03/run-dialyzer-on-elixir-on-travis.html

PR #250, opened 2017-03-07
Deterministically merge FileGroup resolutions

jparise:
This is an incomplete solution to the fact that our existing resolution
merge was non-deterministic with regard to non-qualified names. We now
always take the newer value for a conflicting key, which gets us closer
to the expected behavior.

We unfortunately still aren't implementing the correct scoping rules.
For example, this still allows a non-qualified name to "leak" from one
included file into another. This should be revisited and addressed with
a more complete solution.

### DIFF
--- a/lib/thrift/parser/file_group.ex
+++ b/lib/thrift/parser/file_group.ex
@@ -102,7 +102,12 @@ defmodule Thrift.Parser.FileGroup do
     end)
     |> Map.new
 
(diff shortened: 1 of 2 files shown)


jparise:
@dantswain: Not explicitly, but it's implied by the "reference" implementation. All the spec says is:

> An include makes all the symbols from another file visible (with a prefix) and adds corresponding include statements into the code generated for this Thrift document.

PR #252, opened 2017-03-07
Remove the implicit field identifier warning

jparise:
This is useful information, but we don't currently have a good way to
capture it for later display. As is, printing out these warnings as part
of the parsing path can product a bit of spurious output as part of
normal operations because our mix compile task always parses .thrift
files to determine their build targets.


dantswain:
Would it make sense for this to be optional?

jparise:
> Would it make sense for this to be optional?

Yes, that's related to my vague comment about us not having a good system to print warnings.

I starting playing with the idea of writing a `thrift.lint` mix task that could flag stuff like this, but that's a larger effort unto itself.

PR #253, opened 2017-03-11
Fixed Global resolution

scohen:
We had the concept of global resolution, that is, we would place the
current module's definitions in a global scope. This would have worked,
but it wasn't deterministic.

Now, we tell the file group which module is local, and have it update
its resolutions with only the local definitions when we build their
data. This limits the scope of global variables.

In the near future, I intend to move resolution to the Schema. It makes more sense there and we're effectively scoping resolution to schemas now anyways.

### DIFF
--- a/lib/thrift/generator.ex
+++ b/lib/thrift/generator.ex
@@ -56,8 +56,7 @@ defmodule Thrift.Generator do
 
   def generate_to_string!(%FileGroup{} = file_group) do
     Enum.flat_map(file_group.schemas, fn {_, schema} ->
-      schema
-      |> Map.put(:file_group, file_group)
+      %Schema{schema | file_group: file_group}
       |> generate_schema
     end)
     |> Enum.reverse
@@ -68,6 +67,8 @@ defmodule Thrift.Generator do
   end
 
   def generate_schema(schema) do
+    current_module_file_group = FileGroup.set_current_module(schema.file_group, schema.module)
+    schema = %Schema{schema | file_group: current_module_file_group}
     List.flatten([
       generate_enum_modules(schema),
       generate_const_modules(schema),
--- a/lib/thrift/parser.ex
+++ b/lib/thrift/parser.ex
@@ -71,9 +71,14 @@ defmodule Thrift.Parser do
     |> FileRef.new
     |> ParsedFile.new
 
+    mod_name = file_path
+    |> Path.basename
+    |> Path.rootname
+    |> String.to_atom
+
     FileGroup.new(file_path, normalized_opts)
     |> FileGroup.add(parsed_file)
-    |> FileGroup.update_resolutions
+    |> FileGroup.set_current_module(mod_name)
   end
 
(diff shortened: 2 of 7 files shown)


lib/thrift/generator.ex
```diff
@@ -56,8 +56,7 @@ defmodule Thrift.Generator do
 
   def generate_to_string!(%FileGroup{} = file_group) do
     Enum.flat_map(file_group.schemas, fn {_, schema} ->
-      schema
-      |> Map.put(:file_group, file_group)
+      %Schema{schema | file_group: file_group}
```
jparise: 👍

lib/thrift/parser.ex
```diff
@@ -71,9 +71,14 @@ defmodule Thrift.Parser do
...
 
+    mod_name = file_path
+    |> Path.basename
+    |> Path.rootname
+    |> String.to_atom
+
     FileGroup.new(file_path, normalized_opts)
     |> FileGroup.add(parsed_file)
-    |> FileGroup.update_resolutions
+    |> FileGroup.set_current_module(mod_name)
```
jparise: Should we make the call to `set_current_module` from within `FileGroup.add/1`? It seems like we'd always want to do this whenever we add a new file.

lib/thrift/parser/file_group.ex
```diff
@@ -87,28 +87,29 @@ defmodule Thrift.Parser.FileGroup do
...
-    to_update = file_group.resolutions
-    |> Enum.map(fn {name, v} ->
-      case String.split(Atom.to_string(name), ".") do
-        [_module, enum_name, value_name] ->
-          {:"#{enum_name}.#{value_name}", v}
-        [_initial_module, rest] ->
-          {:"#{rest}", v}
-      end
+
+    current_module = Atom.to_string(module)
```
jparise: We converted the module string _to_ an atom just before we called this function, and here we're converting it back to a string in order to use it. Should it just be a string the entire time?

lib/thrift/parser/file_group.ex
```diff
@@ -87,28 +87,29 @@ defmodule Thrift.Parser.FileGroup do
...
+            [{name, v}, original_mapping]
+        end
     end)
     |> Map.new
 
-    # Merge the non-qualified names into the resolution map. We always replace
-    # existing entries with the same key. This produces a predictable result
-    # but is wrong in the sense that it doesn't implement the expected scoping
-    # rules. For example, this still allows a non-qualified name to "leak"
-    # from one included file into another. TODO: revisit resolution scoping
```
jparise: Thank you for "revisiting". :)

lib/thrift/parser/file_group.ex
scohen: This is one of two cases; in `generator.ex`, we use the schema's module which is an atom. I also felt that the module make sense as an attribute since elixir modules are just atoms.

Also, think of this as step one of two. I think moving all of this to Schema will make so much more sense.

PR #254, opened 2017-03-14
Fixed Global Resolution... again

scohen:
Fixes #253

The problem with the last fix is that generation is recursive. This
means we should be cleaning up each nested context's local lookups when
we move to a deeper context. Otherwise, we're relying on map iteration
order to override a particular value. If the value comes after the one
wer're inserting, we're out of luck.

This PR starts from a known good state with no overrides, and on each
subsequent generation, builds a new resolution map.

### DIFF
--- a/lib/thrift/parser/file_group.ex
+++ b/lib/thrift/parser/file_group.ex
@@ -43,6 +43,7 @@ defmodule Thrift.Parser.FileGroup do
             parsed_files: %{},
             schemas: %{},
             resolutions: %{},
+            immutable_resolutions: %{},
             ns_mappings: %{},
             opts: Keyword.new()
 
@@ -61,6 +62,7 @@ defmodule Thrift.Parser.FileGroup do
     %__MODULE__{file_group |
                 parsed_files: new_parsed_files,
                 schemas: new_schemas,
+                immutable_resolutions: resolutions,
                 resolutions: resolutions}
   end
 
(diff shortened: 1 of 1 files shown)


dantswain:
What sets the order in the first place?  Is it just the order of the file listing or does it come from the order within the file, too?

scohen:
@dantswain , the entries in the resolution map are added to the resolution as they're encountered in the file, and the file is added as its discovered via parsing. 

The problem we were experiencing is the ordering of maps, which from the Erlang docs is arbitrary.

Image we have the following:
```elixir
%{
 "foo.MyEnum", %TypeRef{type: "foo.MyEnum"},
 "bar.MyEnum", %TypeRef{type: "bar.MyEnum"},
 "foo.MyEnum.VAL, %ValRef {type: "MyEnum"},
 "bar.MyEnum.VAL", %ValRef{type: "MyEnum"}
}
```
Now when the first file we encounter adds locals (in this case, `foo`), this becomes:
```elixir
%{
 "foo.MyEnum", %TypeRef{type: "foo.MyEnum"},
 "bar.MyEnum", %TypeRef{type: "bar.MyEnum"},
 "foo.MyEnum.VAL, %ValRef {type: "MyEnum"},
 "bar.MyEnum.VAL", %ValRef{type: "MyEnum"},
 "MyEnum", %TypeRef{type: "foo.MyEnum"},
 "MyEnum.VAL, %ValRef {type: "MyEnum"},
}
```
Now, when we process `bar`, the ordering is important, since we kept the old entries. If we get to the `bar.MyEnum` first via iteration, we'll add it as a local, then, we hit the old local for `MyEnum`, which replaces the original enum. Often, I'd see the local enum reference work, but the enum value reference fail.

Starting with the original map whenever we hit a new context allows us to not have to clean the old local contexts out of the resolution map.

Hope this helps explain it.

scohen:
Also, It's my intention to move resolution out of the FileGroup and into the Schemas. It seems to make sense that schemas should be able to do this local resolution.

PR #255, opened 2017-03-15
Use "Thrift.Generated" as the default namespace

jparise:
Previously, we defaulted to generating our output files in the root, but
that behavior is potentially dangerous because it could class with other
top-level module names.

It's still possible to write files to the root by specifying `""` as the
namespace.

See discussion in #57.

### DIFF
--- a/lib/mix/tasks/compile.thrift.ex
+++ b/lib/mix/tasks/compile.thrift.ex
@@ -41,7 +41,9 @@ defmodule Mix.Tasks.Compile.Thrift do
     config      = Keyword.get(Mix.Project.config, :thrift, [])
     input_files = Keyword.get(config, :files, [])
     output_path = Keyword.get(config, :output_path, "lib")
-    parser_opts = Keyword.take(config, [:include_paths, :namespace])
+    parser_opts =
+      Keyword.take(config, [:include_paths, :namespace])
+      |> Keyword.put_new(:namespace, "Thrift.Generated")
 
     mappings =
       input_files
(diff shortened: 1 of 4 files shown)


lib/thrift/parser.ex
```diff
@@ -91,12 +91,13 @@ defmodule Thrift.Parser do
   #   - convert an atom to a binary and remove the "Elixir." we get from atoms
   #      like `Foo`
   #   - make sure values are valid module names (CamelCase)
-  defp namespace_string(nil), do: nil
+  defp namespace_string(nil), do: "Thrift.Generated"
+  defp namespace_string(""), do: nil
```
dantswain: So they would have to put an empty string to get the root namespace?

What if instead of doing this conversion here you handle the defaults when you parse the arguments in the corresponding mix tasks?  Then you could do something like

```
namespace = if Keyword.has_key?(config, :namespace) do
  Keyword.get(config, :namespace)
else
  "Thrift.Generated"
end
```

That would translate to having "Thrift.Generated" as a default and letting the user override it as `namespace: nil`.

lib/mix/tasks/compile.thrift.ex
```diff
@@ -41,7 +41,9 @@ defmodule Mix.Tasks.Compile.Thrift do
     config      = Keyword.get(Mix.Project.config, :thrift, [])
     input_files = Keyword.get(config, :files, [])
     output_path = Keyword.get(config, :output_path, "lib")
-    parser_opts = Keyword.take(config, [:include_paths, :namespace])
+    parser_opts =
```
dantswain:
```
config
|> Keyword.take ..
|> Keyword.put_new
```
Not that I care that much, just regurgitating what I know credo would say 🤷‍♂️

lib/mix/tasks/compile.thrift.ex
```diff
@@ -27,6 +27,7 @@ defmodule Mix.Tasks.Compile.Thrift do
       search for included files
     * `:namespace` - default namespace for generated modules, which will
       be used when a Thrift file doesn't define its own `elixir` namespace.
+      Defaults to `"Thrift.Generated"`.
```
dantswain: Should we also document the `nil` case?

PR #257, opened 2017-03-27
Constant modules now reuse suitable existing names

jparise:
We were previously writing constants to modules named after their
originating files. This approach was unfortunately too naive because the
resulting module names could conflict with other generated modules that
are spelled the same but different only in letter case.

For example, for a file named `myservice.thrift`:

    const float PI = 3.14
    service MyService {}

We would previously write the constants to `myservice.Myservice`
(`myservice.ex`) and the service definition to `myservice.MyService`
(`my_service.ex`). We now write both to `myservice.MyService` (thanks to
the generator's constants-merging logic) and no longer run into module
name conflicts (see #196).

### DIFF
--- a/lib/thrift/parser/file_group.ex
+++ b/lib/thrift/parser/file_group.ex
@@ -175,12 +175,27 @@ defmodule Thrift.Parser.FileGroup do
   end
 
   def dest_module(file_group, Constant) do
-    # for constants the name of the module is
-    #   <Namespace>.<filename>
-    initial_file = file_group.initial_file
-    base = Path.basename(initial_file, Path.extname(initial_file))
-    name = String.to_atom(base <> "." <> Macro.camelize(base))
-    dest_module(file_group, name)
+    # Default to naming the constants module after the namespaced, camelized
+    # basename of its file. For foo.thrift, this would be `foo.Foo`.
+    base = Path.basename(file_group.initial_file, ".thrift")
+    default = base <> "." <> Macro.camelize(base)
+
+    # However, if we're already going to generate an equivalent module name
+    # (ignoring case), use that instead to avoid generating two modules with
+    # the same spellings but different cases.
+    schema = file_group.schemas[base]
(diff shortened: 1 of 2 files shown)


jparise:
I think my solution is good but not as clean as it could be.

I considered using the Resolver to find candidate names, but that felt like the wrong coupling. I also considered writing my search in terms of recursively `dest_module/2` calls across the schema's symbols, but that felt too complex.

Other suggestions are welcome.

lib/thrift/parser/file_group.ex
```diff
@@ -175,12 +175,27 @@ defmodule Thrift.Parser.FileGroup do
...
 
   def dest_module(file_group, Constant) do
-    # for constants the name of the module is
-    #   <Namespace>.<filename>
-    initial_file = file_group.initial_file
-    base = Path.basename(initial_file, Path.extname(initial_file))
-    name = String.to_atom(base <> "." <> Macro.camelize(base))
-    dest_module(file_group, name)
+    # Default to naming the constants module after the namespaced, camelized
+    # basename of its file. For foo.thrift, this would be `foo.Foo`.
```
scohen: I'm surprised that El/ixir allows a module name that doesn't start with a capital. 
Or, am I misreading something

test/thrift/parser/file_group_test.exs
```diff
@@ -0,0 +1,18 @@
...
+  alias Thrift.Parser.Models.Constant
+
+  test "constant module uses suitable existing name" do
+    with_thrift_files([
+      "myservice.thrift": """
+      const float PI = 3.14
+      service MyService {}
+      """, as: :file_group, parse: "myservice.thrift"]) do
+
+      assert :"Elixir.MyService" == FileGroup.dest_module(file_group, Constant)
```
scohen: I'd like to see some stronger asserts here, for example that the constant is generated in the correct module.

PR #259, opened 2017-04-05
Revise TApplicationException

jparise:
This brings TApplicationException more in line with Elixir's conventions
for creating exceptions. `exception/1` now expects a keyword list with a
:type and an optional :message value. :type can either be an integer or
one of our internal atoms (e.g. :invalid_protocol).

Only the atom form is stored internally. It can be lazily converted back
to an integer at serialization time using `type_id/1`. We don't store
the integer form up-front because it's only used in "exceptional" (ha!)
cases in the server code path.

The list of well-known exception types has been expanded based on the
current Apache Thrift 0.10.0 release (C++ implementation).

Lastly, TApplicationException now has dedicated unit test coverage.

### DIFF
--- a/lib/thrift/exceptions.ex
+++ b/lib/thrift/exceptions.ex
@@ -4,7 +4,7 @@ defmodule Thrift.TApplicationException do
   """
 
   @enforce_keys [:message, :type]
-  defexception message: "", type: :unknown
+  defexception message: "unknown", type: :unknown
 
   # This list represents the set of well-known TApplicationException types.
   # We primarily use their atom names, but we also need their standardized
@@ -28,7 +28,8 @@ defmodule Thrift.TApplicationException do
 
   def exception(args) when is_list(args) do
     type = Keyword.fetch!(args, :type) |> normalize_type
-    %__MODULE__{message: args[:message] || "", type: type}
+    message = args[:message] || Atom.to_string(type)
+    %__MODULE__{message: message, type: type}
   end
 
   @doc """
--- a/lib/thrift/exceptions.ex
+++ b/lib/thrift/exceptions.ex
@@ -4,7 +4,7 @@ defmodule Thrift.TApplicationException do
   """
 
   @enforce_keys [:message, :type]
-  defexception message: nil, type: :unknown
+  defexception message: "", type: :unknown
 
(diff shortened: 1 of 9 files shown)


lib/thrift/protocol/binary.ex
```diff
@@ -111,7 +108,7 @@ defmodule Thrift.Protocol.Binary do
   end
 
   def deserialize(:application_exception, binary) when is_binary(binary) do
-    do_read_application_exception(binary, %TApplicationException{})
+    do_read_application_exception(binary, TApplicationException.exception(type: :unknown))
```
fishcakez: I think this would be cleaner to accumulate keyword and then pass that to `TApplicationException.exception/1` on `@stop`.

lib/thrift/exceptions.ex
```diff
@@ -1,38 +1,47 @@
...
+    unsupported_client_type:  10,
+    loadshedding:             11,
+    timeout:                  12,
+    injected_failure:         13,
+  ]
 
-  for {atom_name, type_id} <- @exception_mappings do
-    def exception_type_to_int(unquote(atom_name)), do: unquote(type_id)
+  def exception(args) when is_list(args) do
+    %__MODULE__{message: args[:message], type: normalize_type(args[:type])}
```
fishcakez: To be clear message MUST be a string.

jparise:
@fishcakez, thanks for the `do_read_application_exception/2` suggestion. I had intended to revisit that logic but forgot before pushing this PR. I definitely prefer your approach.

With regard to exception's value `:message`, due to the way Thrift's `TApplicationException` (wire-level) struct is defined, I believe that value is technically optional, in which case we need to support a `nil` value in our exception struct.

fishcakez:
@jparise then need to define `message/1` to set string message when `nil`.

jparise:
@fishcakez I just went with `""` (empty string) for the `nil` case because I don't think there's another reasonable default.

The only other thought I had was to generate per-type default messages, which could be useful, but they would basically be repeated the information represented by the atom names.

fishcakez:
@jparise perhaps converting the atom (or integer?) to a string?

jparise:
@fishcakez, yes, I think that's better.

PR #262, opened 2017-04-06
Fix line numbers for thrift tests

### LINKED ISSUE #229: `thrift_test` loses test line numbers
I don't know if it's easy to fix this and it shouldn't hold us back I don't think, but I thought I'd make an issue for discussion's sake (maybe someone has an idea for a quick fix).  The `thrift_test` macro makes it impossible to run, e.g., `mix test test/thrift/generator/binary_protocol_test.exs:599` because the line numbers get lost.  Maybe it's as simple as adding `location: :keep` to the `defmacro`?

fishcakez:
Create/register tests inline in test module so that test line numbers can be used with `mix test`.

Closes #229.

### DIFF
--- a/test/support/lib/thrift_test_case.ex
+++ b/test/support/lib/thrift_test_case.ex
@@ -76,7 +76,7 @@ defmodule ThriftTestCase do
   end
 
   defp alias?(module, modules) do
-    # alias when module when parent in namespace does not exist
+    # alias when parent module in namespace does not exist
     not Enum.any?(modules, &(:lists.prefix(&1, module) and &1 != module))
   end
 
--- a/test/support/lib/thrift_test_case.ex
+++ b/test/support/lib/thrift_test_case.ex
@@ -70,8 +70,8 @@ defmodule ThriftTestCase do
       nil ->
         []
       modules ->
-        modules = Enum.map(modules, fn({module, _}) -> Module.split(module) end)
-        Enum.filter_map(modules, &alias?(&1, modules), &Module.concat/1)
+        parts = Enum.map(modules, fn({module, _}) -> Module.split(module) end)
+        Enum.filter_map(parts, &alias?(&1, parts), &Module.concat/1)
       end
   end
 
--- a/test/support/lib/thrift_test_case.ex
+++ b/test/support/lib/thrift_test_case.ex
@@ -163,7 +163,7 @@ defmodule ThriftTestCase do
   end
 
   defp loaded?(source_file \\ nil, mod) do
-    source_file = source_file && String.to_char_list(source_file)
+    source_file = source_file && String.to_charlist(source_file)
     case :code.is_loaded(mod) do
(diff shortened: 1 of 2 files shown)


test/support/lib/thrift_test_case.ex
```diff
@@ -1,131 +1,134 @@
...
+    filename
+  end
+
+  defp require_file(file, dir) do
+    case Code.require_file(file, dir) do
+      nil ->
+        []
+      modules ->
+        modules = Enum.map(modules, fn({module, _}) -> Module.split(module) end)
+        Enum.filter_map(modules, &alias?(&1, modules), &Module.concat/1)
```
jparise: Or (shorter and doesn't rebind `modules`):
```elixir
modules
|> Enum.map(fn({module, _}) -> Module.split(module) end)
|> Enum.filter_map(&alias?(&1, modules), &Module.concat/1)
```

test/support/lib/thrift_test_case.ex
```diff
@@ -142,22 +145,52 @@ defmodule ThriftTestCase do
...
+    module = erlang_module(source_file)
+    if loaded?(source_file, module) do
+      module
+    else
+      erlang_compile(source_file)
+    end
+  end
+
+  defp loaded?(source_file \\ nil, mod) do
+    source_file = source_file && String.to_char_list(source_file)
```
jparise: Isn't the `String.to_charlist/1` spelling preferred for Elixir 1.3+?

jparise:
One related comment is that the Erlang bits are currently useful (for integration testing), but I hope that we won't need them after too long once we're standing firmly on our own legs.

For binary compatibility testing, I think we should generate some "golden" test data files and write them as both serialized Thrift data and their equivalent Erlang term format. That will allow us to validate our serialization/deserialization code without using any of the Erlang runtime.

test/support/lib/thrift_test_case.ex
```diff
@@ -1,131 +1,134 @@
...
+      nil ->
+        []
+      modules ->
+        modules = Enum.map(modules, fn({module, _}) -> Module.split(module) end)
+        Enum.filter_map(modules, &alias?(&1, modules), &Module.concat/1)
+      end
+  end
+
+  defp alias?(module, modules) do
+    # alias when module when parent in namespace does not exist
```
fishcakez: That's supposed to be `when parent module`

test/support/lib/thrift_test_case.ex
fishcakez: That isn't quite equivalent, but it just shows how confusing that piece of code is. We need the rebound `modules` as second argument to `alias?/2`.

test/support/lib/thrift_test_case.ex
fishcakez: Yes it is, didn't update from the prior code.

PR #263, opened 2017-04-07
Clean up service test shutdown

jparise:
Also, remove some unused variables.

### DIFF
--- a/test/thrift/generator/service_test.exs
+++ b/test/thrift/generator/service_test.exs
@@ -370,8 +370,11 @@ defmodule Thrift.Generator.ServiceTest do
 
   thrift_test "clients exit if they try to use a closed client", ctx do
     Process.flag(:trap_exit, true)
+
+    ref = Process.monitor(ctx.server)
     Process.exit(ctx.server, :normal)
 
+    assert_receive {:DOWN, ^ref, _, _, _}
     assert {{:error, :econnrefused}, _} = catch_exit(Client.friend_ids_of(ctx.client, 1234))
   end
 
--- a/test/thrift/generator/service_test.exs
+++ b/test/thrift/generator/service_test.exs
@@ -82,7 +82,7 @@ defmodule Thrift.Generator.ServiceTest do
   alias Thrift.Generator.ServiceTest.SimpleService.Binary.Framed.Client
   alias Thrift.Generator.ServiceTest.UsernameTakenException
 
-  def start_server(port, recv_timeout \\ :infinity) do
+  defp start_server(port, recv_timeout \\ :infinity) do
     :thrift_socket_server.start(
       handler: ServerSpy,
       port: port,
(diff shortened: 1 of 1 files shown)


test/thrift/generator/service_test.exs
```diff
@@ -370,8 +370,6 @@ defmodule Thrift.Generator.ServiceTest do
 
   thrift_test "clients exit if they try to use a closed client", ctx do
     Process.flag(:trap_exit, true)
-
-    ref = Process.monitor(ctx.server)
```
fishcakez: I think should wait for the `:DOWN` of the server before doing the client call. Otherwise the server may not have shut down before the client request. This may solve the occasional failure: https://travis-ci.org/pinterest/elixir-thrift/jobs/219331581#L484. However `ranch` will asynchronously close the listen socket/client sockets so it may not stop this error completely - the `gen_tcp` sockets are closed asynchronously when their owning process exits and `ranch` only waits for the owning processes to exit.

Note that `Process.exit(pid, :normal)` is a bit of an antipattern because it will only stop a child process that is trapping exits and adheres to OTP principles. Rather should use `Mod.stop` or if not provided`:sys.terminate` for OTP processes. `Process.exit(pid, :normal)` is unlikely to work on any non-OTP process. Therefore the only real use case for this is to stop a child Supervisor asynchronously in a test - as we are doing here but we actually want synchronous stop.

jparise:
@fishcakez, I think I've incorporated both bits of your feedback. I still see the occasional test failure, so I'm not sure if the `assert_receive` timeout should be increased or if there's something incorrect in the code itself.

dantswain:
Haven't had a chance to look at this deeply, but sometimes sporadic failures with async code like this can be caused by side effects of previous tests.  Do you get sporadic failures if you run just the flakey test?

fishcakez:
@jparise oh this doesn't use `ranch`, it uses a custom acceptor pool: https://github.com/apache/thrift/blob/ae971ce917bf9b60ee8ae83b834dad1eb149a82f/lib/erl/src/thrift_socket_server.erl#L276. It seems that stop is a cast, so it can return before the server has stopped. The library has not been kept up to date to use `:gen_server.stop` (synchronous and guarantees process has exited). There is a `stop_server/1` call that will do the equivalent (it does `mod.stop` + monitor) defined early in the test module. However this would also be equivalent to monitor + `Process.exit(pid, :normal)`.

test/thrift/generator/service_test.exs
scohen: @fishcakez , I've only taken a cursory look, but isn't the server a `thrift_socket_server` instance here? Ranch shouldn't be involved.

fishcakez: @scohen you are correct: https://github.com/pinterest/elixir-thrift/pull/263#issuecomment-292737382

jparise:
@fishcakez / @scohen, how does this look now?

fishcakez:
@jparise it looks good. Does this resolve the race condition? I think we could also use that in the `setup` `on_exit` call?

We should also remove all of the `unlink` calls. We shouldn't be unlinking because a crash will orphan the unlinked child, instead trap exits if need to handle abnormal exit - it seems like trapping is already happening.

jparise:
@fishcakez, I wasn't sure about using `stop_server/1` in `on_exit` because it contain an `assert`, but it seems to work well, and it cleans up the code a bit more.

I've also removed the `unlink` calls.

I haven't seen a test failure with these latest changes, but it wasn't happening very frequently before, so we may need to wait and see.

test/thrift/generator/service_test.exs
```diff
@@ -113,14 +111,7 @@ defmodule Thrift.Generator.ServiceTest do
...
       [handler_pid, client, server]
-      |> Enum.each(fn pid ->
-        ref = Process.monitor(pid)
-        Process.exit(pid, :normal)
-        receive do
-          {:DOWN, ^ref, _, _, _} ->
-            :ok
-        end
-      end)
+      |> Enum.each(&stop_server/1)
```
fishcakez: `stop_server/1` should only work on the `server`?

PR #275, opened 2017-09-05
Typedefs in other modules weren't working properly

### LINKED ISSUE #268: Included enums not being resolved correctly
Updating to the most recent thrift_tng commit breaks some of our existing (working) code.  The following test case is distilled down from something in our working thrift IDL files, so if the test is incorrect then I've done a bad job trying to simplify.

```
  @thrift_file name: "numbers.thrift", contents: """
  enum Number {
    ZERO = 0,
    ONE = 1,
    TWO = 2,
    THREE = 3
  }
  """

  @thrift_file name: "includes_enum.thrift", contents: """
  include "numbers.thrift"

  struct Person {
    1: optional Number favoriteNumber = Number.ZERO
  }
  """

  thrift_test "including an emum" do
    assert %Person{farvorite_number: 0} == %Person{}
  end
```

Results in

```
elixir-thrift - thrift_tng! ❯ mix test test/thrift/generator/binary_protocol_test.exs
Excluding tags: [pending: true]

** (RuntimeError) Fatal error: Could not find value: Number.ZERO
    (thrift) lib/thrift/generator/utils.ex:279: Thrift.Generator.Utils.resolution_failed/1
    (thrift) lib/thrift/generator/utils.ex:186: Thrift.Generator.Utils.quote_value/3
    (thrift) lib/thrift/generator/struct_generator.ex:16: anonymous fn/2 in Thrift.Generator.StructGenerator.generate/4
    (elixir) lib/enum.ex:1229: Enum."-map/2-lists^map/1-0-"/2
    (thrift) lib/thrift/generator/struct_generator.ex:14: Thrift.Generator.StructGenerator.generate/4
    (thrift) lib/thrift/generator.ex:181: anonymous fn/3 in Thrift.Generator.generate_struct_modules/1
    (stdlib) lists.erl:1263: :lists.foldl/3
    (thrift) lib/thrift/generator.ex:179: Thrift.Generator.generate_struct_modules/1
    (thrift) lib/thrift/generator.ex:75: Thrift.Generator.generate_schema/1
    (thrift) lib/thrift/generator.ex:52: anonymous fn/3 in Thrift.Generator.generate!/2
    (elixir) lib/enum.ex:958: anonymous fn/3 in Enum.flat_map/2
    (stdlib) lists.erl:1263: :lists.foldl/3
    (elixir) lib/enum.ex:1772: Enum.flat_map/2
    (elixir) lib/enum.ex:966: Enum.flat_map_list/2
    (elixir) lib/enum.ex:967: Enum.flat_map_list/2
    (thrift) test/support/lib/thrift_test_case.ex:97: ThriftTestCase.generate_elixir_files/3
    (thrift) test/support/lib/thrift_test_case.ex:84: ThriftTestCase.generate_files/4
    (thrift) test/support/lib/thrift_test_case.ex:36: ThriftTestCase.quoted_contents/2
    test/thrift/generator/binary_protocol_test.exs:658: (module)

```

scohen:
Due to the way typedefs work (they're different than everything else),
resolution wasn't working properly on them, this commit adds proper
resolution so you can define them in remote modules.

Fixes #268

### DIFF
--- a/lib/thrift.ex
+++ b/lib/thrift.ex
@@ -22,6 +22,10 @@ defmodule Thrift do
   @typedoc "Thrift message types"
   @type message_type :: :call | :reply | :exception | :oneway
 
+  @doc """
+  Returns a list of atoms, each of which is a name of a Thrift primitive type.
+  """
+  @spec primitive_names() :: [Thrift.Parser.Types.Primitive.t]
   def primitive_names do
     [:bool, :i8, :i16, :i32, :i64, :binary, :double, :byte, :string]
   end
--- a/lib/thrift.ex
+++ b/lib/thrift.ex
@@ -10,7 +10,7 @@ defmodule Thrift do
 
   @typedoc "Thrift data types"
   @type data_type ::
-    :bool | :byte | :i16 | :i32 | :i64 | :double | :string |
+    :bool | :byte | :i8 | :i16 | :i32 | :i64 | :double | :string | :binary |
     {:map, data_type, data_type} | {:set, data_type} | {:list, data_type}
 
   @type i8   :: (-128..127)
@@ -21,4 +21,8 @@ defmodule Thrift do
 
   @typedoc "Thrift message types"
   @type message_type :: :call | :reply | :exception | :oneway
+
+  def primitive_names do
+    [:bool, :i8, :i16, :i32, :i64, :binary, :double, :byte, :string]
+  end
 end
--- a/lib/thrift/parser/models.ex
+++ b/lib/thrift/parser/models.ex
@@ -470,11 +470,11 @@ defmodule Thrift.Parser.Models do
     end
 
     defp merge(schema, %TEnum{} = enum) do
-      %Schema{schema | enums: put_new_strict(schema.enums, enum.name, canonicalize_name(schema, enum))}
+      %Schema{schema | enums: put_new_strict(schema.enums, enum.name, canonicalize_name(schema.module, enum))}
     end
 
     defp merge(schema, %Exception{} = exc) do
-      fixed_fields = schema
+      fixed_fields = schema.module
       |> canonicalize_name(exc)
       |> canonicalize_fields()
 
@@ -482,7 +482,7 @@ defmodule Thrift.Parser.Models do
     end
 
     defp merge(schema, %Struct{} = s) do
-      fixed_fields = schema
+      fixed_fields = schema.module
       |> canonicalize_name(s)
       |> canonicalize_fields()
 
@@ -490,7 +490,7 @@ defmodule Thrift.Parser.Models do
     end
 
     defp merge(schema, %Union{} = union) do
-      fixed_fields = schema
+      fixed_fields = schema.module
       |> canonicalize_name(union)
       |> canonicalize_fields()
 
@@ -498,19 +498,19 @@ defmodule Thrift.Parser.Models do
     end
 
     defp merge(schema, %Service{} = service) do
-      %Schema{schema | services: put_new_strict(schema.services, service.name, canonicalize_name(schema, service))}
+      %Schema{schema | services: put_new_strict(schema.services, service.name, canonicalize_name(schema.module, service))}
     end
 
     defp merge(schema, {:typedef, actual_type, type_alias}) do
       %Schema{schema | typedefs: put_new_strict(schema.typedefs, atomify(type_alias), canonicalize_type(schema.module, actual_type))}
     end
 
-    defp canonicalize_name(%{module: nil}, model) do
+    defp canonicalize_name(nil, model) do
       model
     end
 
-    defp canonicalize_name(schema, %{name: name} = model) do
-      %{model | name: :"#{schema.module}.#{name}"}
+    defp canonicalize_name(module, %{name: name} = model) do
+      %{model | name: canonicalize_type(module, name)}
     end
 
     defp put_new_strict(map, key, value) do
@@ -545,7 +545,7 @@ defmodule Thrift.Parser.Models do
       |> String.split(".")
 
       case split_type_name do
-        [^module | rest] ->
+        [^module | _rest] ->
           # this case accounts for types that already have the current module in them
           type_name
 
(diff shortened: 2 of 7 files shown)


lib/thrift/parser/models.ex
```diff
@@ -475,6 +475,30 @@ defmodule Thrift.Parser.Models do
...
+    end
+    for type <- [:bool, :i8, :i16, :i32, :i64, :binary, :string, :byte] do
+      defp canonicalize_type(_, unquote(type)) do
+        unquote(type)
+      end
+    end
+    defp canonicalize_type(schema, type_name) when is_atom(type_name) do
+      :"#{schema.module}.#{type_name}"
+    end
+    defp canonicalize_name(schema, type_name) when is_bitstring(type_name) do
```
dantswain: Is this a typo?  Should it be `canonicalize_type` or should this be up with the other `canonicalize_name` clauses?

scohen: I think i can delete that one.

scohen:
@dantswain Nothing is obvious; I'd appreciate a test that test case.
I'm going to guess that it's looping in the `FileGroup.resolve()` step.

dantswain:
OK, I don't have a repro yet but I have isolated where the loop is happening.  It's calling `FileGroup.resolve` with a value - in our case `:"BrowserType.ALL"` which comes from a `browser_type.thrift` file.  It doesn't find a resolution for that in the `resolutions` map, so it ends up calling `resolve(%FileGroup{}, nil)` and that's triggering the "path" clause because `is_atom(nil)` is true.  THAT ends up calling itself again because `resolutions[nil]` is ALSO `nil`.

scohen:
Interesting; now I'm wondering why it can't resolve `:"BrowserType.ALL"`. That's an enum, correct?

dantswain:
Yeah, it is an enum.

scohen:
No repro yet?

dantswain:
Not yet.  I'm at ElixirConf so free time is an issue :/

scohen:
You bastard! Have fun!

lib/thrift/parser/models.ex
```diff
@@ -475,6 +475,27 @@ defmodule Thrift.Parser.Models do
...
+    defp canonicalize_type(schema, {:set, elem_type}) do
+      {:set, canonicalize_type(schema, elem_type)}
+    end
+    defp canonicalize_type(schema, {:list, elem_type}) do
+      {:list, canonicalize_type(schema, elem_type)}
+    end
+    defp canonicalize_type(schema, {:map, {key_type, val_type}}) do
+      {:map, {canonicalize_type(schema, key_type), canonicalize_type(schema, val_type)}}
+    end
+    for type <- [:bool, :i8, :i16, :i32, :i64, :binary, :string, :byte] do
```
jparise: Missing `:byte` and `:double`?

dantswain:
@scohen I was able to reproduce by modifying your test case a little (you might say by a _hair_):

```
  @thrift_file name: "additions.thrift", contents: """
  enum ChocolateAdditionsType {
    ALMONDS = 1,
    NOUGAT  = 2,
    HAIR = 3
  }

  typedef set<ChocolateAdditionsType> ChocolateAdditions
  """

  @thrift_file name: "chocolate.thrift", contents: """
  include "additions.thrift"

  struct Chocolate {
    1: optional additions.ChocolateAdditions extra_stuff
    2: optional additions.ChocolateAdditionsType secret_ingredient = ChocolateAdditionsType.HAIR
  }
  """
```

It no longer hangs if the value is `additions.ChocolateAdditionsType.HAIR` (i.e., namespaced with the file name).  You could argue that it is a bit strange that the type needs to be namespaced but the value does not, and I might agree, but this would pass through the `thrift` compiler unscathed :/

jparise:
I wonder if the type used by the field's default value also needs to be "canonicalized".

```elixir
    @spec new(integer, boolean, Types.t, charlist, Literals.t) :: t
    def new(id, required, type, name, default) do
      %Field{id: id,
             type: type,
             name: atomic_snake(name),
             required: required,
             default: cast(type, default)}
    end
```

lib/thrift/parser/models.ex
scohen: `:byte` is at the end; `:double` is missed.

scohen:
@jparise can we bump the line limit to > 80? That seems a little short.

dantswain:
SO CLOSE.  I no longer get a hang with the latest commit, but now I'm getting an error resolving a constant that _is_ properly namespaced.  It looks like the code is duplicating the filename part of the namespace - i.e., if I have something like

```
include "foo.thrift"

struct Thing {
  1: optional foo.Foo foo_kind = foo.Foo.FOO_OF_THE_FIRST_KIND
}
```

I'm getting an error that it can't find a value for `foo.foo.Foo.FOO_OF_THE_FIRST_KIND` - i.e., it's duplicating the first `foo`.

I'll try to come up with a replicating test case if you don't have something off the top of your head.

scohen:
@dantswain that's probably because I'm resolving the namespaces rather stupidly; This is easy to repro with the latest test case.

lib/thrift/parser/models.ex
```diff
@@ -470,35 +470,135 @@ defmodule Thrift.Parser.Models do
...
+    defp add_namespace_to_type(module, {:set, elem_type}) do
+      {:set, add_namespace_to_type(module, elem_type)}
+    end
+    defp add_namespace_to_type(module, {:list, elem_type}) do
+      {:list, add_namespace_to_type(module, elem_type)}
+    end
+    defp add_namespace_to_type(module, {:map, {key_type, val_type}}) do
+      {:map, {add_namespace_to_type(module, key_type), add_namespace_to_type(module, val_type)}}
+    end
+    for type <- [:bool, :i8, :i16, :i32, :i64, :binary, :string, :byte] do
```
jparise: Still missing `:double` here?

dantswain:
WOOHOO It works for me now!

joshuawscott:
Works for my case as well from #268 Thanks!

lib/thrift.ex
```diff
@@ -21,4 +21,8 @@ defmodule Thrift do
 
   @typedoc "Thrift message types"
   @type message_type :: :call | :reply | :exception | :oneway
+
+  def primitive_names do
```
jparise: Could you add a short `@doc` for this one given that it's in the top-level module?

PR #276, opened 2017-09-08
Add support for Thrift annotations

### LINKED ISSUE #270: Support Thrift IDL annotation syntax
Annotations are written within parentheses:

```thrift
struct Test {
    1: string foo (is_red = "yes");
    2: string bar
}
```

jparise:
Annotations appear within parentheses as part of various Thrift IDL
types such as fields and structures. They can be used to hint the
compiler or inform runtime behavior; the specification is pretty
open-ended.

This change adds support for parsing annotations (whose presence
previously resulted in parse errors) and storing them on our parser
models.

We currently support annotations on:

 - Enums
 - Fields
 - Functions
 - Services
 - Structures, Unions, and Exceptions

We can parse annotations on typedefs but don't currently store them
anywhere due to the way we internally represent a schema's typedefs as a
simple map of strings to types. We can revisit that later, perhaps as
part of a broader typedef refactoring.

Fixes #270

### DIFF
--- a/lib/thrift/parser.ex
+++ b/lib/thrift/parser.ex
@@ -9,6 +9,9 @@ defmodule Thrift.Parser do
   @typedoc "A Thrift IDL line number"
   @type line :: pos_integer | nil
 
+  @typedoc "A map of Thrift IDL annotations"
+  @type annotations :: %{String.t => String.t}
+
   @typedoc "A schema path element"
   @type path_element :: String.t | atom
 
--- a/lib/thrift/parser.ex
+++ b/lib/thrift/parser.ex
@@ -9,7 +9,7 @@ defmodule Thrift.Parser do
   @typedoc "A Thrift IDL line number"
   @type line :: pos_integer | nil
 
-  @typedoc "A map of Thrift IDL annotations"
+  @typedoc "A map of Thrift annotation keys to values"
   @type annotations :: %{String.t => String.t}
 
   @typedoc "A schema path element"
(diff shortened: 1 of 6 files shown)


lib/thrift/parser.ex
```diff
@@ -9,6 +9,9 @@ defmodule Thrift.Parser do
   @typedoc "A Thrift IDL line number"
   @type line :: pos_integer | nil
 
+  @typedoc "A map of Thrift IDL annotations"
```
dantswain: After looking at the .thrift file, it's a little more clear what "a map of Thrift IDL annotations" means, but at first this was confusing (a map from what to what?).  Maybe something like "A map of thrift annotation keys to values"?  Are they always in that format?

jparise:
> Will there be a follow-up PR to use these? Right now it seems like they would basically just be lost once the generated code is written to file, so it would only really be useful for someone using the parser directly to generate their own code.

I don't have anything specific planned. The immediate motivation is recognizing these tokens in the grammar so we don't blow up on `.thrift` files that include annotations. I figured it would be useful to add basic support for capturing them while I was there, even if they're just thrown away after the code generation step.

Some ideas have been floated before around hinting the code generator to do things like generating lists for `set` types (instead of the "heavier" `MapSet`), and annotations could be a good way to allow for those things. (There were better ideas than that; I just can't recall them right now.)

dantswain:
Makes sense to me.

PR #278, opened 2017-09-12
Check erlang thrift files generated

fishcakez:
We should check that the thrift generator is returning a successful exit status to prevent hidden errors generating erlang source files.


fishcakez:
It seems that the failing tests was an issue with thrift in travis and now its fixed on their side once we are checking the result.

PR #279, opened 2017-09-14
Merge 'thrift_tng' into master

jparise:
This change represents a full merge of out `thrift_tng` branch into master.

This code is sufficiently stable to warrant working on it in master, although
it isn't yet ready for a full 2.0 release.


dantswain:
I'm stuck without #275 being resolved :/  I would only be able to use a release on about 2/3 of our projects without that.

I also feel like we could do with a pass through the documentation?

If the goal here is just to merge to master and hammer out the rest of these things against that, I'm OK with that.  If the goal is to do a release right away, then I'm a bit hesitant..

jparise:
The goal is the second thing, and I'm going to make sure that open PR can be retargeted to master after this merge.

jparise:
Er, I mean the first thing. Just get us back onto a single branch. We won't release until it's all in good shape.

PR #280, opened 2017-09-19
Adopt the Ecto Credo configuration

jparise:
This is a more liberal set of linter rules. I've made a few small
changes, but this is largely the same configuration that Ecto uses.

### DIFF
--- a/.credo.exs
+++ b/.credo.exs
@@ -1,130 +1,79 @@
-%{
-  configs: [
-    %{
-      #
-      # Run any config using `mix credo -C <name>`. If no config name is given
-      # "default" is used.
-      name: "default",
-      #
-      # These are the files included in the analysis:
-      files: %{
-        #
-        # You can give explicit globs or simply directories.
-        # In the latter case `**/*.{ex,exs}` will be used.
-        included: ["lib/", "src/"],
-        excluded: [~r"/_build/", ~r"/deps/"]
-      },
-      #
-      # If you create your own checks, you must specify the source files for
-      # them here, so they can be loaded by Credo before running the analysis.
-      requires: [],
-      #
-      # Credo automatically checks for updates, like e.g. Hex does.
-      # You can disable this behaviour below:
-      check_for_updates: false,
-      #
-      # If you want to enforce a style guide and need a more traditional linting
-      # experience, you can change `strict` to `true` below:
-      strict: false,
-      #
-      # If you want to use uncolored output by default, you can change `color`
-      # to `false` below:
-      color: true,
-      #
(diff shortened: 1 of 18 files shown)


.credo.exs
```diff
@@ -1,130 +1,79 @@
...
+      {Credo.Check.Readability.FunctionNames},
+      {Credo.Check.Readability.LargeNumbers, false},
+      {Credo.Check.Readability.MaxLineLength, false},
+      {Credo.Check.Readability.ModuleAttributeNames},
+      {Credo.Check.Readability.ModuleDoc},
+      {Credo.Check.Readability.ModuleNames},
+      {Credo.Check.Readability.ParenthesesOnZeroArityDefs, false},
+      {Credo.Check.Readability.ParenthesesInCondition},
+      {Credo.Check.Readability.PredicateFunctionNames},
+      {Credo.Check.Readability.SinglePipe, false}, # Off for now
```
fishcakez: I would like this to be on because I think single pipe (`var |> call()`) is much less clear than `call(var)`. Is this off for now because we are using it? Perhaps we can turn it on to force no-single pipe just for changes?

.credo.exs
```diff
@@ -1,130 +1,79 @@
...
-        {Credo.Check.Readability.StringSigils},
-        {Credo.Check.Readability.TrailingBlankLine},
-        {Credo.Check.Readability.TrailingWhiteSpace},
-        {Credo.Check.Readability.VariableNames},
-        {Credo.Check.Readability.Semicolons},
-        {Credo.Check.Readability.SpaceAfterCommas},
+      {Credo.Check.Refactor.ABCSize, false},
+      {Credo.Check.Refactor.CondStatements},
+      {Credo.Check.Refactor.DoubleBooleanNegation, false},
+      {Credo.Check.Refactor.FunctionArity, max_arity: 8},
```
fishcakez: 8 is still many arguments, can we reduce this to 6? I think Ecto must have a much bigger number 😊 .

.credo.exs
jparise: I should have improved the comment. It turned it off because there are many places we need to fix up, and I thought that would be nicer as a separate change. I'll do it as part of this PR now though.

jparise:
Let's move forward with this in its current form. I agree that it could be useful to document the things we have disabled thus far, but I don't know all of the rationale behind Ecto's individual choices.

I added a note to one rule that we've specifically disabled because it crashed Credo. We should investigate that separately and open an issue on Credo for it.

fishcakez:
@jparise looks great. I see that was a much bigger change required than I expected...

PR #282, opened 2017-09-21
Delegate response handling to client implementation

jparise:
This reduces the amount of generated code by doing the bulk of the
response processing in the common client implementation module.

It also highlights an oddity of the way we handle `void` service calls.
Even though those calls are successful, they run through the "error"
path because we don't currently have a way to differentiate between
`nil` (void) and `nil` (missing) "success" field values in the unpacked
response structure. "Reply" messages are only supposed to have either a
"success" field or an exception representation, so we'll need to revisit
our reply implementation before we can improve the success path here.

### DIFF
--- a/lib/thrift/binary/framed/client.ex
+++ b/lib/thrift/binary/framed/client.ex
@@ -202,37 +202,32 @@ defmodule Thrift.Binary.Framed.Client do
 
     case Connection.call(conn, {:call, rpc_name, serialized_args, tcp_opts}, gen_server_timeout) do
       {:ok, data} ->
-        deserialize_response(data, deserialize_module)
+        data
+        |> deserialize_module.deserialize
+        |> unpack_response
 
       {:error, _} = err ->
         err
     end
   end
 
-  defp deserialize_response(data, deserialize_module) do
-    case deserialize_module.deserialize(data) do
-      {%{success: nil} = response, ""} ->
-        responses = response
-        |> Map.from_struct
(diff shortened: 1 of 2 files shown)


fishcakez:
I think think this improves the control flow but we could improve it further by reducing the amount of generated code. We aren't actually doing anything that requires metaprogramming but the module concatenation is nice to have. Perhaps we could clean this module up by removing as much generation as possible and delegating to the client implementation for shared logic. For example:
```elixir
deserialize_module = Module.concat(response_module, :BinaryProtocol)
quote bind_quoted: [module: module, rpc_name: rpc_name] do
  ClientImpl.call(client, rpc_name, deserialize_module, serialized_args, opts)
end
```

jparise:
Yes, that's a good observation. I'll have a look at moving the bulk of the logic into the client implementation module.

lib/thrift/binary/framed/client.ex
```diff
@@ -188,19 +188,50 @@ defmodule Thrift.Binary.Framed.Client do
...
+
+      {:error, _} = err ->
+        err
+    end
+  end
+
+  defp deserialize_response(data, deserialize_module) do
+    case deserialize_module.deserialize(data) do
+      {%{success: nil} = response, ""} ->
+        responses = response
```
fishcakez: Can we move the nested cases (and perhaps pipe) into their own functions to split this up?

PR #283, opened 2017-09-22
Improve unpack_response/1 test coverage

jparise:
This also includes additional commentary on the implementation as well
as a small optimization for the case where a void function doesn't throw
any exceptions.

### DIFF
--- a/test/thrift/generator/service_test.exs
+++ b/test/thrift/generator/service_test.exs
@@ -14,6 +14,8 @@ defmodule Thrift.Generator.ServiceTest do
 
   service SimpleService {
     bool ping(),
+    void check_username(1: string username)
+      throws(1: UsernameTakenException taken),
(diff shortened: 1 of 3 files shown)


test/thrift/generator/service_test.exs
```diff
@@ -251,6 +253,18 @@ defmodule Thrift.Generator.ServiceTest do
     assert {:ok, nil} == Client.mark_inactive(ctx.client, 5529)
   end
 
+  thrift_test "it handles void functions that can throw exceptions", ctx do
+    ServerSpy.set_reply(:noreply)
+    assert {:ok, nil} == Client.check_username(ctx.client, "foo")
+  end
+
+  thrift_test "it handles void functions that do throw exceptions", ctx do
+    ServerSpy.set_reply({:exception, {:UsernameTakenException, 'blowie up'}})
```
fishcakez: 💣

PR #286, opened 2017-10-03
Upgrade out Docker image to Thrift 0.10.0

jparise:
This is only used by Travis CI to power our integration tests.


dantswain:
What's the motivation for changing it?  Do we support anything new in 0.10.0, or were there any important bug fixes?

jparise:
We've switched to 0.10.0 internally at Pinterest so it's the version most folks have on their machines (and also the default from `brew install thrift`).

There were some good Erlang improvements in 0.10.0, but they're not relevant for this project now that we no longer ship any Thrift-generated code or runtime libraries.

If we want to keep our integration tests on 0.9.3, we should consider adding a runtime version check to enforce that expectation.

dantswain:
I'm sort of drawing a blank.  Are we only using the thrift binary to generate the Erlang code for reference testing?  If that's the case then I suppose this change is fine.

jparise:
Right, we're only using the `thrift` binary in the unit test paths for integration / reference testing.

PR #287, opened 2017-10-03
Update Travis build matrix

dantswain:
* Test a couple versions of Elixir/OTP
* Only generate a coverage report for the most recent versions
* Only build master branch (still build PRs)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,17 +1,31 @@
 language: elixir
-elixir: 1.3.4
-otp_release: 18.2
+
+branches:
+  only:
+    - "master"
+
 matrix:
   include:
-    - elixir: 1.5.0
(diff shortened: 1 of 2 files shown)


jparise:
Could we do this without the additional script by conditionally setting a `TEST_COMMAND` variable within `.travis.yml` and then using that as `mix $TEST_COMMAND`?

dantswain:
We could.  It just seems easier to me to have an actual script.

jparise:
Looks good to me. The only tradeoff is seeing the variable `COVERALLS=(true|false)` versions `TEST_COMMAND=(test|coveralls.travis)` in the Travis matrix UI, but they're both pretty clear.

I like this more now that the docker bits have moved into the test script, too.

Issue #261, opened 2017-04-06
Review binary framed client TCP error handling


fishcakez:
Regarding this error handling: https://github.com/pinterest/elixir-thrift/blob/1168b29d4099045fb788f489472cf56fa75ca962/lib/thrift/binary/framed/client.ex#L229-L236

I believe that a timeout will cause the client to get out of sync and can only ever return an prior sequence id - if it does not keep timing out. This occurs because the next `call` will receive the result of the prior call - if it receives anything. This can continue forever with no hope of recovery.

I have not tried this but it appears that by returning an error atom on tcp error it will try to raise an atom later on https://github.com/pinterest/elixir-thrift/blob/1168b29d4099045fb788f489472cf56fa75ca962/lib/thrift/generator/binary/framed/client.ex#L85, which will fail. Should that use `:inet.format_error/1` on the `:inet` error atom to raise a RuntimeError or create a custom exception?

fishcakez:
We need to introduce a custom exception, something like `ThriftConnectionError` to handle the connection errors: https://github.com/pinterest/elixir-thrift/blob/89130e838a70d7f1fb68e7da3b9b25acb6ac86af/lib/thrift/generator/binary/framed/client.ex#L85-L86 If we use the bang variants then this will fail with an argument error because we are attempting to raise `{:error, atom}`. Instead it should be something like:
```elixir
{:error, reason} ->
  raise ThriftConnectionError, reason: reason
```

The `reason` can also be `:timeout`: https://github.com/pinterest/elixir-thrift/blob/89130e838a70d7f1fb68e7da3b9b25acb6ac86af/lib/thrift/binary/framed/client.ex#L252-L253 In this situation we also want to close the socket because its possible for the following infinite loop of errors to occur:

1. Send call with seq id X.
2. Timeout receiving the response for call with seq id X.
3. Send call with seq id X + 1.
4. Receive response for seq id X when expecting X + 1, therefore raise `TApplicationException`.
5. Send call with seq id X + 2.
6. Receive response for seq id X + 1 when expecting X + 2, therefore raise `TApplicationException`.

It might even be that 4. or 6. etc will also timeout because the server likely blocks until the prior the call finishes. Therefore X+1 has to wait for X to finish before it is started, and is then much more likely to also timeout.

Likely we also want to have a different exception for a timeout as a user may want to handle `:timeout`, differently to a connection error, perhaps `ThriftTimeoutError`.

jparise:
[...]

PR #291, opened 2017-10-04
Increase default assert receive timeout

### LINKED ISSUE #235: Intermittent test failures
Travis tests fail maybe 30% of the time (guesstimate) and it's always some kind of timeout that goes away with a rebuild.  @scohen I _think_ this is in the service / behaviour code and smells like a synchronization problem.  If you have some idea what might be failing, I have some tricks up my sleeve to deal with that kind of thing.

fishcakez:
CI can be slow so that the default (100ms) is not always enough.

Fixes #235

### DIFF
--- a/test/test_helper.exs
+++ b/test/test_helper.exs
@@ -1,5 +1,4 @@
-ExUnit.configure(exclude: [pending: true])
-ExUnit.start()
+ExUnit.start(exclude: [pending: true], assert_receive_timeout: 500)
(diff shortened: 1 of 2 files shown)


fishcakez:
This fix doesn't resolve #235. #235 has different tcp error reason so increasing the timeout would not match. `:econnreset` v `:econnrefused`. We should handle this as part of #261.

PR #292, opened 2017-10-04
Increase assert_received timeout to 200ms

### LINKED ISSUE #235: Intermittent test failures

jparise:
We occasionally exceed the 100ms timeout when run within Travis CI's
environment so double the timeout to 200ms to avoid spurious test
failures.

Fixes #235

### DIFF
--- a/test/thrift/generator/service_test.exs
+++ b/test/thrift/generator/service_test.exs
@@ -419,7 +419,7 @@ defmodule Thrift.Generator.ServiceTest do
 
     stop_server(ctx.server)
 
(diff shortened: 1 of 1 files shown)


jparise:
Closing in favor of #291

PR #293, opened 2017-10-06
Misc doc tweaks 1

dantswain:
I started to do a big pass through the docs and got sidetracked, so I decided to just PR what I could when I could to avoid conflicts.  Plus this way I get some of that sweet hacktoberfest cred.

* Improve the mix configuration examples in README and mix tasks.
* Remove the source_ref from the docs config since we've merged to master.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,10 +79,20 @@ In this example, we gather all of the `.thrift` files under the `thrift`
 directory and write our output to the `lib/thrift/` directory:
 
 ```elixir
-thrift: [
-  files: Path.wildcard("thrift/**/*.thrift"),
-  output_path: "lib/thrift/"
-]
+# example mix.exs
+defmodule MyProject.Mixfile do
+  # ...
+
+  def project do
+    [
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -79,10 +79,20 @@ In this example, we gather all of the `.thrift` files under the `thrift`
...
-  files: Path.wildcard("thrift/**/*.thrift"),
-  output_path: "lib/thrift/"
-]
+# example mix.exs
+defmodule MyProject.Mixfile do
+  # ...
+
+  def project do
+    [
+      # other srettings...
```
jparise: "settings" here and in the two other places below.

mix.exs
```diff
@@ -51,8 +51,8 @@ defmodule Thrift.Mixfile do
      docs: [
        main: "README",
        extras: ["README.md": [title: "README"]],
-       source_ref: "thrift_tng",
```
jparise: 👍

Issue #294, opened 2017-10-16
Enumerations that use reserved keywords can generate uncompilable code


thecodeboss:
This thrift schema:
```thrift
enum Operator {
  AND = 0
}
```
generates an `Operator` module containing this invalid code:
```elixir
defmacro(and) do
  0
end
```
`and` is a reserved keyword in Elixir and hence we get this compile error:
```elixir
== Compilation error in file lib/thrift/generated/operator.ex ==
** (SyntaxError) lib/thrift/generated/operator.ex:3: syntax error before: 'and'
    (elixir) lib/kernel/parallel_compiler.ex:121: anonymous fn/4 in Kernel.ParallelCompiler.spawn_compilers/1
```

nichochar:
Great job @thecodeboss !

PR #295, opened 2017-10-17
Handle reserved keywords

thecodeboss:
[Similarly to apache/thrift](https://github.com/apache/thrift/blob/19baeefd8c38d62085891d7956349601f79448b3/compiler/cpp/src/thrift/thriftl.ll#L273), we disallow keywords being used in the Thrift IDL.

However the keywords restriction only applies to exact-casing, so for example we must support "AND" as a valid enum. When down-casing this value, it conflicts with the Elixir keyword "and", and so using some macro magic we can get around this limitation.

Additionally, since we down-case method names too, similar logic applies when generating a client. The server implementation did not require any changes.

Fixes: #294

### DIFF
--- a/lib/thrift/generator/binary/framed/client.ex
+++ b/lib/thrift/generator/binary/framed/client.ex
@@ -41,6 +41,9 @@ defmodule Thrift.Generator.Binary.Framed.Client do
     bang_name = :"#{underscored_name}!"
     options_bang_name = :"#{underscored_options_name}!"
 
+    underscored_name = Macro.pipe(underscored_name, quote do unquote end, 0)
+    bang_name = Macro.pipe(bang_name, quote do unquote end, 0)
+
     vars = Enum.map(function.params, &Macro.var(&1.name, nil))
 
     assignments = function.params
--- a/test/mix/tasks/thrift.generate_test.exs
+++ b/test/mix/tasks/thrift.generate_test.exs
@@ -4,9 +4,14 @@ defmodule Mix.Tasks.Thrift.GenerateTest do
 
   @project_root Path.expand("../../../", __DIR__)
   @fixture_project Path.join(@project_root, "test/fixtures/app")
+  @temp_dir Path.join([@fixture_project, "tmp"])
 
   setup do
-    on_exit(fn -> File.rm_rf!(Path.join(@fixture_project, "lib")) end)
+    File.mkdir_p!(@temp_dir)
+    on_exit(fn ->
+      File.rm_rf!(Path.join(@fixture_project, "lib"))
+      File.rm_rf!(@temp_dir)
+    end)
     :ok
   end
 
@@ -38,8 +43,16 @@ defmodule Mix.Tasks.Thrift.GenerateTest do
 
   test "specifying an invalid Thrift file" do
     in_fixture fn ->
-      assert_raise Mix.Error, ~r/reserved language keyword/, fn ->
-        run([__ENV__.file])
+      bad_schema = """
+      struct InvalidTest {
+        1: i32 cannotDoArithmeticInThrift = 1 + 1,
+      }
+      """
+      bad_file = Path.join([@temp_dir, "invalid.thrift"])
+      File.write!(bad_file, bad_schema)
+
+      assert_raise Mix.Error, ~r/Error parsing/, fn ->
+        run([bad_file])
       end
     end
   end
--- a/test/mix/tasks/thrift.generate_test.exs
+++ b/test/mix/tasks/thrift.generate_test.exs
@@ -38,7 +38,7 @@ defmodule Mix.Tasks.Thrift.GenerateTest do
 
   test "specifying an invalid Thrift file" do
     in_fixture fn ->
-      assert_raise Mix.Error, ~r/Error parsing/, fn ->
+      assert_raise Mix.Error, ~r/reserved language keyword/, fn ->
         run([__ENV__.file])
       end
     end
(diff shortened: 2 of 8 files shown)


test/thrift/parser/parser_test.exs
```diff
@@ -648,6 +648,29 @@ defmodule Thrift.Parser.ParserTest do
...
+    struct continue {}
+    """
+
+    expected_error = "line 1: cannot use reserved language keyword \"continue\""
+    assert_raise RuntimeError, expected_error, fn ->
+      parse(thrift)
+    end
+  end
+
+  test "names can be reserved keywords if they have a difference case" do
```
jparise: `s/difference/difference/`

test/mix/tasks/thrift.generate_test.exs
```diff
@@ -38,7 +38,7 @@ defmodule Mix.Tasks.Thrift.GenerateTest do
 
   test "specifying an invalid Thrift file" do
     in_fixture fn ->
-      assert_raise Mix.Error, ~r/Error parsing/, fn ->
+      assert_raise Mix.Error, ~r/reserved language keyword/, fn ->
```
jparise: I think we should still have a test case that covers the "Error parsing" error.

lib/thrift/generator/binary/framed/client.ex
```diff
@@ -41,6 +41,9 @@ defmodule Thrift.Generator.Binary.Framed.Client do
     bang_name = :"#{underscored_name}!"
     options_bang_name = :"#{underscored_options_name}!"
 
+    underscored_name = Macro.pipe(underscored_name, quote do unquote end, 0)
```
jparise: @fishcakez, do you have any other suggestions on how to write this that involve a bit less macro magic?

fishcakez: The standard way to solve this would be to use `bind_quoted`. However that requires binding variables, which makes the code in the generated modules much less clear:
```elixir
underscored_name = :the_fun_name
def unquote(underscored_name)(...) do
```
The more bound variables, the less clear the generated code becomes when using `Macro.to_string`.
 
Therefore we want to get around binding variables in the generated code. This means we need to include an `unquote/1` call in the AST which has a literal atom argument, and not a variable. If we use `unquote(var)` or `bind_quoted` we won't get a literal atom but a variable binding followed by an unquote call with variable argument. As `quote/1` is part of `Kernel.SpecialForms` we are limited in how we can do this. The easiest way I can think of to avoid creating the AST by hand is to use `quote/0` (which is not a Kernel.SpecialForm) and pipe into it with this function call.

thecodeboss:
> Because we’re now an Elixir-native generator, I wonder if we should only check for words that would conflict with Elixir names. The only reason to complain about “volatile” is because of C++, for example.
That being said, it’s nice to be helpful.

@jparise Thrift is supposed to be used across multiple languages, so I think it's a good idea to conform to the standard and not allow any of those keywords. Otherwise someone could end up adding a keyword to their thrift schema, only to have to change it later once they discover it won't compile for one of the other languages.

lib/thrift/generator/binary/framed/client.ex
thecodeboss: I had come up with a few approaches for this, but after consulting with @fishcakez he was able to find `Macro.pipe` to solve it, and we agreed it seemed the cleanest.
I'd say the next cleanest approach is:
```elixir
underscored_name = Code.string_to_quoted!(" unquote(:#{underscored_name})() ")
```
The string shows exactly the code that ends up getting generated, namely:
```elixir
defmacro unquote(:and)() do
  ...
end
```
But converting from a string to a quoted expression is not as good as just building the quoted expression directly with `Macro.pipe`, though I think both approaches are good ones.

@jparise do you have thoughts on what may work better here?

test/thrift/parser/parser_test.exs
nichochar: s/difference/different/

jparise: I fixed it. :)

lib/thrift/generator/binary/framed/client.ex
jparise: I think what you have is fine. I was just fishing (ha!) for something more obvious.

fishcakez: An alternative is to delay some of the metaprogramming to the generated file. So that we generate a file that calls macros. For example the file might look like:
```elixir
import Thift.Binary.Framed.Client.DSL

thrift_rpc :literal_underscored_name, ...
```
Of course we would lose the significant advantage of being able to see the explicit calls.

jparise:
> Thrift is supposed to be used across multiple languages, so I think it's a good idea to conform to the standard and not allow any of those keywords. Otherwise someone could end up adding a keyword to their thrift schema, only to have to change it later once they discover it won't compile for one of the other languages.

I think that's a good perspective. There's an ongoing cost to keeping up with the "official" set of reserved words, but that's not unreasonable.

test/mix/tasks/thrift.generate_test.exs
thecodeboss: That's fair. This particular error is already covered by one of the other test cases, so I've updated the diff with an "Error parsing" case

test/thrift/parser/parser_test.exs
thecodeboss: Me too

PR #299, opened 2017-11-27
Support multiple destination module name casings

### LINKED ISSUE #297: Compilation error when service name is lowercase.
service test
{
    string hello()
}

(CompileError) lib/gen-ex/:"_elixir/thrift/generated/test".ex:91: invalid alias: "Elixir.Thrift.Generated.test().HelloArgs.BinaryProtocol".

jparise:
Previously, we expected the "struct name" portion of the input name to
be a suitable module name string (for Module.concat/1). Lowercased names
failed this assumption because they're not valid aliases.

This change reworks the destination module name processing to ensure
that the "struct name" string starts with an initial capitalize letter.

Fixes #297

### DIFF
--- a/lib/thrift/parser/file_group.ex
+++ b/lib/thrift/parser/file_group.ex
@@ -213,14 +213,13 @@ defmodule Thrift.Parser.FileGroup do
 
     case file_group.ns_mappings[module_name] do
       nil ->
-        Module.concat(List.flatten([Elixir, struct_name]))
+        Module.concat([struct_name])
       namespace = %Namespace{} ->
-        namespace_module = namespace.path
-        |> String.split(".")
-        |> Enum.map(&Macro.camelize/1)
-        |> Enum.join(".")
-        |> String.to_atom
-        Module.concat(List.flatten([namespace_module, struct_name]))
+        namespace_parts =
+          namespace.path
+          |> String.split(".")
+          |> Enum.map(&Macro.camelize/1)
+        Module.concat(namespace_parts ++ [struct_name])
     end
   end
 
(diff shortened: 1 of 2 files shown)


nichochar:
👍

lib/thrift/parser/file_group.ex
```diff
@@ -203,12 +203,15 @@ defmodule Thrift.Parser.FileGroup do
...
+    name_parts =
+      name
+      |> Atom.to_string
+      |> String.split(".", parts: 2)
 
-    case file_group.ns_mappings[thrift_module] do
+    module_name = name_parts |> Enum.at(0) |> String.to_atom
+    struct_name = name_parts |> Enum.at(1) |> initialcase
+
+    case file_group.ns_mappings[module_name] do
```
fishcakez: Could we clean up the rest of the module name creation too because its quite complex compared to what it needs to be.

Without a namespace we can wrap the string in a list to create the alias: `Module.concat([struct_name])`, don't need to add `Elixir`. I think its actually an anti-pattern because `Elixir` is really `:"Elixir.Elixir"`. Fortunately concat will strip all leading `"Elixir."`.

With a namespace we don't need to `Enum.join` or `String.to_atom`. Instead a list of binaries with an alias prepended: `Module.concat([namespace_module | struct_name])`.

jparise: Those are good suggestions. I've applied them in 7afef2a.

For your last note, I think it needs to be `Module.concat([namespace] ++ [struct_name])`, however, because the namespace is the list and the struct name is the binary.

fishcakez: Yes good point. I think I was going by the old data structures for that part.

PR #303, opened 2017-12-23
Remove the client-side retry logic

jparise:
In addition to the moderate state-management complexity introduced by
this feature, we've also found that implementing message-level retry
logic directly within a Thrift client to not be ideal. Instead, retries
are better implemented by either application logic (which has complete
understanding of the message payloads, idempotency, etc.) or closer to
the wire level, which can make stronger guarantees about whether the
original message was or was not delivered.

See #261

### DIFF
--- a/lib/thrift/binary/framed/client.ex
+++ b/lib/thrift/binary/framed/client.ex
@@ -118,7 +118,7 @@ defmodule Thrift.Binary.Framed.Client do
   end
 
   @doc false
-  def disconnect(info, %{sock: sock} = s) do
+  def disconnect(info, %{sock: sock}) do
     :ok = :gen_tcp.close(sock)
 
     case info do
@@ -126,14 +126,14 @@ defmodule Thrift.Binary.Framed.Client do
         Connection.reply(from, :ok)
         {:stop, :normal, nil}
 
-      {:error, :closed} ->
+      {:error, :closed} = error ->
         Logger.error("Connection closed")
-        {:connect, info, %{s | sock: nil}}
+        {:stop, error, nil}
 
-      {:error, reason} ->
+      {:error, reason} = error ->
         reason = :inet.format_error(reason)
         Logger.error("Connection error: #{reason}")
-        {:connect, info, %{s | sock: nil}}
+        {:stop, error, nil}
     end
   end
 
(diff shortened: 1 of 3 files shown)


lib/thrift/binary/framed/client.ex
```diff
@@ -252,8 +211,8 @@ defmodule Thrift.Binary.Framed.Client do
       {:error, :timeout} = timeout ->
         {:reply, timeout, s}
 
-      {:error, :closed} = err ->
-        {:disconnect, {:retry, err}, %{s |last_message: {call_args, caller}}}
+      {:error, :closed} = error ->
+        {:disconnect, error, s}
```
fishcakez: We should rely on next clause to handle this, as we need to reply as well and would be identical behavior.

fishcakez: Can we also do the same for : timeout above? The logic doesn't seem right there...

jparise: I had generalized it by some tests failed. They seemed sensitive to the 3- vs. 4- element version of `:disconnect`.

I didn’t look further for this particular change but agree there’s still room for improvement.

jparise: I just took another pass at this, and there's something about the `{:error, :closed}` case that appears to be different than the other errors. The current tests only work if we defer the reply until `disconnect/2` sends its `:stop`; if we reply immediately, something goes amiss and the process don't exit.

lib/thrift/binary/framed/client.ex
```diff
@@ -252,8 +211,8 @@ defmodule Thrift.Binary.Framed.Client do
       {:error, :timeout} = timeout ->
         {:reply, timeout, s}
 
-      {:error, :closed} = err ->
-        {:disconnect, {:retry, err}, %{s |last_message: {call_args, caller}}}
```
fishcakez: `{:error, :closed}` used to be special case so that it would always trigger a reconnect attempt - however if `retry: false` (default) the process is stopped immediately after setting up a fresh connection https://github.com/pinterest/elixir-thrift/pull/303/files#diff-217ed11585237b96d356fba460274021L144 and a reply would never be sent. I think this is a bug in the test/previous behavior.

lib/thrift/binary/framed/client.ex
fishcakez: Do you mean that the call does not get an exit here: https://github.com/pinterest/elixir-thrift/pull/303/files#diff-095d95cc2c507ddb3650fc4f4887facaR355? I think that it is fine for the call not to get an exit there, and instead wait for the `:EXIT` from the client:
```elixir
assert Client.friend_ids_of(client, 1234) == {:error, :closed}
assert_receive {:EXIT, ^client, {:error, :closed}}
```

PR #306, opened 2017-12-31
Timeouts should also disconnect to close socket

jparise:
See #261

### DIFF
--- a/lib/thrift/binary/framed/client.ex
+++ b/lib/thrift/binary/framed/client.ex
@@ -113,7 +113,7 @@ defmodule Thrift.Binary.Framed.Client do
   end
 
   @doc false
-  def disconnect(info, %{sock: sock}) do
+  def disconnect(info, %{sock: sock, timeout: timeout}) do
     :ok = :gen_tcp.close(sock)
 
     case info do
@@ -123,6 +123,10 @@ defmodule Thrift.Binary.Framed.Client do
 
       {:error, :closed} = error ->
         Logger.error("Connection closed")
+        {:stop, error, nil}
+
+      {:error, :timeout} = error ->
+        Logger.error("Connection timed out (#{timeout}ms)")
         {:stop, error, nil}
 
       {:error, reason} = error ->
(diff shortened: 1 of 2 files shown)


lib/thrift/binary/framed/client.ex
```diff
@@ -125,6 +125,10 @@ defmodule Thrift.Binary.Framed.Client do
         Logger.error("Connection closed")
         {:stop, error, nil}
 
+      {:error, :timeout} = error ->
+        Logger.error("Connection timed out (#{timeout}ms)")
```
fishcakez: The `timeout` is the default timeout, which won't be `1` in the test below. We could do something like

```elixir
assert ExUnit.CaptureLog.capture_log(fn ->
  assert {:error, :timeout} = Client.friend_ids_of_with_options(client, 12_914, [tcp_opts: [timeout: 1]])
  assert_receive {:EXIT, ^client, {:error, :timeout}}
 end) =~ "1ms"
```

test/thrift/generator/service_test.exs
```diff
@@ -310,10 +310,12 @@ defmodule Thrift.Generator.ServiceTest do
     )
   end
 
-  thrift_test "it has a configurable socket timeout", ctx do
+  thrift_test "it has a configurable socket timeout", %{client: client} do
+    Process.flag(:trap_exit, true)
     ServerSpy.set_reply({:sleep, 1000, [1, 3, 4]})
 
-    assert {:error, :timeout} = Client.friend_ids_of_with_options(ctx.client, 12_914, [tcp_opts: [timeout: 1]])
+    assert {:error, :timeout} = Client.friend_ids_of_with_options(client, 12_914, [tcp_opts: [timeout: 1]])
```
fishcakez: ^ No default timeout is set just for that call

lib/thrift/binary/framed/client.ex
jparise: Ah, you're right, and because the recv timeout option can be set on a per-call basis, that value isn't available in `disconnect/2` unless we included it as a new term in the error tuple (e.g. `{:error, :timeout, 1}`). Is that a standard approach?

Alternatively, I could log from `handle_call/2` or not log the timeout value at all.

fishcakez: I think that logging the timeout length is very useful. Usually the value is passed to disconnect callback and logged there, so any method similar to you suggested (append to the tuple) is good.

PR #307, opened 2017-12-31
Use any free port for our test server process

jparise:
Passing 0 as the port tells the TCP subsystem to select any free port,
and we can querying the server's listening port via a {:get, :port}
call.

This lets us drop the random_port() helper function and makes things
simpler overall.

### DIFF
--- a/test/thrift/generator/service_test.exs
+++ b/test/thrift/generator/service_test.exs
@@ -92,10 +92,11 @@ defmodule Thrift.Generator.ServiceTest do
       handler: ServerSpy,
       socket_opts: [recv_timeout: recv_timeout],
       framed: true]
-    with {:ok, pid} <- :thrift_socket_server.start(opts),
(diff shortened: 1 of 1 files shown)


test/thrift/generator/service_test.exs
```diff
@@ -85,13 +85,20 @@ defmodule Thrift.Generator.ServiceTest do
...
+  defp start_server(recv_timeout \\ :infinity) do
+    opts = [
+      port: 0,
       service: :simple_service_thrift,
-      socket_opts: [recv_timeout: recv_timeout])
+      handler: ServerSpy,
+      socket_opts: [recv_timeout: recv_timeout],
+      framed: true]
+    with {:ok, pid} <- :thrift_socket_server.start(opts),
+         port <- GenServer.call(pid, {:get, :port}) do
```
fishcakez: Its strange to use a with `<-` when always matching, should use a `=` instead.

PR #308, opened 2018-01-01
Remove with _with_options function variants

jparise:
These were introduced to avoid potential conflicts between our options
parameter (named `opts`) and an `opts` function paramter named by the
Thrift IDL.

This resulted in doubling the number of generated methods (as well as
another level of function call indirection), in addition to making the
resulting user-callable API a little bit more complex.

Instead, go back to using just two generated functions (the "regular"
call and the "bang!" exception-raising variant) and rename our options
parameter to `rpc_opts`, which is very unlikely to conflict in practice.

### DIFF
--- a/lib/thrift/generator/binary/framed/client.ex
+++ b/lib/thrift/generator/binary/framed/client.ex
@@ -29,19 +29,20 @@ defmodule Thrift.Generator.Binary.Framed.Client do
 
   defp generate_handler_function(service_module, function) do
     args_module = Module.concat(service_module, Service.module_name(function, :args))
-
     response_module = Service.module_name(function, :response)
+    rpc_name = Atom.to_string(function.name)
 
-    underscored_name = function.name
+    # Make two Elixir-friendly function names: an underscored version of the
+    # Thrift function name and a "bang!" exception-raising variant.
+    function_name = function.name
     |> Atom.to_string
     |> Macro.underscore
     |> String.to_atom
+    bang_name = :"#{function_name}!"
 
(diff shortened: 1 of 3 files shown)


jparise:
I think we could also consider writing typespecs for these generated functions. That would trigger dializer warnings if the arguments became mismatched.

lib/thrift/generator/binary/framed/client.ex
```diff
@@ -29,19 +29,20 @@ defmodule Thrift.Generator.Binary.Framed.Client do
...
+    bang_name = :"#{function_name}!"
 
-    underscored_options_name = :"#{underscored_name}_with_options"
-    bang_name = :"#{underscored_name}!"
-    options_bang_name = :"#{underscored_options_name}!"
-
-    underscored_name = Macro.pipe(underscored_name, quote do unquote end, 0)
+    # Apply some macro magic to the names to avoid conflicts with Elixir
+    # reserved symbols like "and".
+    function_name = Macro.pipe(function_name, quote do unquote end, 0)
```
fishcakez: This hack is to get the double unquote in the quoted code without using bind quoted? Bind quoted would generate slightly less nice code when outputted as string because it would assign function names to a variable and then unquote that in the definition. I can't think of a cleaner hack off the top of my head as this should really be using `bind_quoted`. The language doesn't really cater for quoted -> string.

jparise: This was introduced in #295. It would be nice to improve this bit in a follow-up change; then we wouldn't need any "magic" comment at all here. 😉

PR #313, opened 2018-01-09
Use atoms to represent enums

### LINKED ISSUE #312: Use atoms for enums
Currently Enum values are represented by integers in Elixir.

```thrift
enum Color { RED = 1, GREEN = 2, BLUE = 3 }

struct Pixel {
  1: optional Color color
}
```

```elixir
require Color
assert Color.RED == 1
pixel = %Pixel{color: Color.RED}

# Is equivalent to:
pixel = %Pixel{color: 1}

case pixel.color do
  Color.RED -> :ok
end
```

It seems like it would be nicer if enums were represented by atoms.

```elixir
pixel = %Pixel{color: :RED}

case pixel.color do
  :RED -> :ok
end
```

# Pros

1. If you inspect/print a value, it's confusing to see a number and have to go find the thrift definition and map the number back into a name that you can reason about. An atom is directly readable.

```elixir
> IO.inspect user
%User{status: 7}           # what is that?
```
...versus...
```
> IO.inspect user
%User{status: :INACTIVE}   # OK makes sense
```

2. Less generated code, faster compile times.

3. You don't have to require a generated enum module to use the values.

4. It's more Elixir-y, I think.

# Cons

1. No longer catches invalid enum values. However it didn't catch enums being used in the wrong context, so we need Dialyzer annotations either way for that.

pguillory:
Previously, deserializing an enum value would yield an integer. Since you would
not want to hard code these integers into your Elixir code, and because they're
essentially unreadable, we generated macros/functions to convert back and forth
between enum names and values.

With this commit, enum values are deserialized as atoms, which is more
idiomatic and people should generally be able to work with directly.

We still generate the macros, which should provide some backwards
compatibility. They generate a deprecation warning when used. At some point in
the future we should remove them completely.

The atoms used match the enum definitions exactly. We no longer need to worry
about manipulating the case and adding underscores to generate valid Elixir
identifiers.

Closes #312.

### DIFF
--- a/lib/thrift/generator/enum_generator.ex
+++ b/lib/thrift/generator/enum_generator.ex
@@ -1,46 +1,37 @@
 defmodule Thrift.Generator.EnumGenerator do
 
   def generate(name, enum) do
-    macro_defs = Enum.map(enum.values, fn {key, value} ->
+    macro_defs = Enum.map(enum.values, fn {key, _value} ->
       macro_name = key
       |> to_name
       |> Macro.pipe(quote do unquote end, 0)
 
       quote do
-        defmacro unquote(macro_name)(), do: unquote(value)
+        defmacro unquote(macro_name)() do
+          IO.warn "Thrift enum macros are deprecated. " <>
+              "See https://github.com/pinterest/elixir-thrift/issues/312."
+          unquote(key)
+        end
       end
     end)
 
-    member_defs = Enum.map(enum.values, fn {_key, value} ->
+    member_defs = Enum.map(enum.values, fn {key, _value} ->
       quote do
-        def member?(unquote(value)), do: true
-      end
-    end)
-
-    name_member_defs = Enum.map(enum.values, fn {key, _value} ->
-      enum_name = to_name(key)
-      quote do
-        def name?(unquote(enum_name)), do: true
+        def member?(unquote(key)), do: true
       end
     end)
 
     value_to_name_defs = Enum.map(enum.values, fn {key, value} ->
-      enum_name = to_name(key)
       quote do
-        def value_to_name(unquote(value)), do: {:ok, unquote(enum_name)}
+        def value_to_name(unquote(value)), do: {:ok, unquote(key)}
       end
     end)
 
     name_to_value_defs = Enum.map(enum.values, fn {key, value} ->
-      enum_name = to_name(key)
       quote do
-        def name_to_value(unquote(enum_name)), do: {:ok, unquote(value)}
+        def name_to_value(unquote(key)), do: {:ok, unquote(value)}
       end
     end)
-
-    names = enum.values
-    |> Keyword.keys
-    |> Enum.map(&to_name/1)
 
     quote do
       defmodule unquote(name) do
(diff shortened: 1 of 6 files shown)


lib/thrift/generator/enum_generator.ex
```diff
@@ -1,47 +1,38 @@
 defmodule Thrift.Generator.EnumGenerator do
 
   def generate(name, enum) do
-    macro_defs = Enum.map(enum.values, fn {key, value} ->
+    macro_defs = Enum.map(enum.values, fn {key, _value} ->
```
jparise: I've been trying to use comprehensions a bit more for cases like this. What do you think?

```elixir
macro_defs = for {key, _value} <- enum.values do
  # ...
end
```

jparise (commented):
I don't feel strongly about backwards compatibility here. I think it's helpful but should be short-lived (i.e. not in the 2.0 release). Because everyone is tracking this repo directly from git right now, we don't need to worry about semver guarantees yet.

dantswain:
This would be a breaking change for me.  I commented on #312.  Sorry I missed the issue when you originally posted it :/

test/thrift/generator/models_test.exs
```diff
@@ -21,71 +21,60 @@ defmodule Thrift.Generator.ModelsTest do
...
-    assert Operator.and == 0
-    assert Operator.member?(0) == true
-    assert Operator.name?(:and) == true
-    assert Operator.value_to_name(0) == {:ok, :and}
-    assert Operator.value_to_name!(0) == :and
-    assert Operator.name_to_value(:and) == {:ok, 0}
-    assert Operator.name_to_value!(:and) == 0
-    assert Operator.meta(:names) == [:and]
+    assert Operator.member?(:AND) == true
+    assert Operator.value_to_name(0) == {:ok, :AND}
```
jparise: This is unrelated to this change, I'm still not a fan of the `value_to_enum` / `name_to_value` function names. I think I'd prefer just `name/1` and `value/1` with no bang! variants, but I admit that I don't work with enum values as often as you and @dantswain.

dantswain (changes requested):
I'm going to :x: only because I'd really like to get some discussion around this before it gets merged.  I apologize for missing the initial discussion :(

lib/thrift/generator/enum_generator.ex
pguillory: Oh, actually I've been trying to avoid them when the match can fail. The `for` behavior of silently returning an empty list has bitten me a few times. It's not fun to debug. So I've come to prefer the way `Enum.map` raises an error, even though it's a little more verbose.

EEx templates are an exception. They basically rely on `for` loops and it's not as big a deal when elements are missing from an HTML page.

It's not a strong preference though. I don't mind using the other syntax if you prefer it.

pguillory:
@dantswain No problem, I didn't wait long before putting up code but there's no rush.

lib/thrift/generator/enum_generator.ex
fishcakez: The `for` issue should only be a problem when using data from outside the libraries control, which is not the case here. Here we know that this will be a 2-tuple because we are supplying the arguments.

fishcakez (approved):
The implementation is good but should await discussion in RFC issue.

test/thrift/generator/models_test.exs
pguillory: Sounds good to me.

PR #316, opened 2018-01-18
Test under Elixir 1.6

jparise:


.travis.yml
```diff
@@ -8,9 +8,9 @@ matrix:
   include:
     - elixir: 1.4.5
       otp_release: 19.3
```
dantswain: We should probably also still test 1.5.x?

fishcakez:
Ah we use an internal function that needed to change arity to optional support greek locale that has different rules.

.travis.yml
jparise: In the past, we've just tested the oldest and newer supported version, but I don't feel strongly. I'll re-add a matrix entry for 1.5.

PR #317, opened 2018-01-18
Avoid using String.Casing.titlecase_once/1

jparise:
This is a private Elixir library function, and its arity changed to
`/2` in Elixir 1.6 to support a new `mode` parameter.

Instead, we can just perform the "upcase" transformation ourselves if we
assume the names only contain ASCII characters, which is reasonable in
this context.


fishcakez:
I think this would be clearer and technically allow supporting more than ascii range:
```elixir
{first, rest} = String.next_grapheme(string)
String.upcase(first) <> rest
```

PR #320, opened 2018-01-30
Raise ConnectionError for bang function errors

### LINKED ISSUE #334: Bug in server code
I believe the `raise` call here will not work, since there is a data structure mismatch.

https://github.com/pinterest/elixir-thrift/blob/5de8597ca961496001b2c504d23af56442cbd6bb/lib/thrift/generator/binary/framed/client.ex#L72

jparise:
The previous code attempted to raise an error atom as an exception.
Instead, wrap that error reason in a new ConnectionError exception.

See #261, fixes #334


lib/thrift/exceptions.ex
```diff
@@ -47,6 +47,15 @@ defmodule Thrift.TApplicationException do
   defp normalize_type(type) when is_integer(type), do: :unknown
 end
 
+defmodule Thrift.ConnectionError do
+  @enforce_keys [:reason]
+  defexception [:reason]
+
+  def message(%{reason: reason}) do
+    "Connection error: #{:inet.format_error(reason)} (#{reason})"
```
pguillory: Might want to special case `:closed` and `:timeout`. They come out of `Thrift.Binary.Framed.Client` but are not POSIX errors.

PR #321, opened 2018-01-30
Improve log handling in tests and binary server

fishcakez:


fishcakez:
@jparise we could do yes. I did this PR to show the bugs with the flaky tests. We can add metadata at another time.

PR #322, opened 2018-01-30
Wait for all server processes to exit on server stop

fishcakez:


fishcakez:
We can close this due to #350 .

PR #324, opened 2018-02-06
Fix gen_server_opts not being passed to GenServer

### LINKED ISSUE #172: Client does not accept name as an option
Currently it is not possible to do
`Client.start_link("localhost", 33722, [name: __MODULE__])`

thecodeboss:
In the docs, the `gen_server_opts` option was listed, but not actually hooked up to pass the opts to the `GenServer` handlers.
This fixes #172, and the tests I added verify that.

### DIFF
--- a/lib/thrift/binary/framed/client.ex
+++ b/lib/thrift/binary/framed/client.ex
@@ -81,15 +81,13 @@ defmodule Thrift.Binary.Framed.Client do
 
      - `send_timeout`: An integer that governs how long our connection waits when sending data.
 
-     - `gen_server_opts`: A keyword list of options that control the gen_server behaviour.
-
-     - `timeout`:  The amount of time to wait (in milliseconds) for a reply from a `GenServer` call.
-
+  Additionally, the options `:name`, `:debug`, and `:spawn_opt`, if specified, will be passed to
+  the underlying `GenServer`. See `GenServer.start_link/3` for details on these options.
   """
   @spec start_link(String.t, (0..65_535), options) :: GenServer.on_start
   def start_link(host, port, opts) do
-    {gen_server_opts, opts} = Keyword.pop(opts, :gen_server_opts, [])
-    Connection.start_link(__MODULE__, {host, port, opts}, gen_server_opts)
+    {gen_server_opts, client_opts} = Keyword.split(opts, [:debug, :name, :spawn_opt])
+    Connection.start_link(__MODULE__, {host, port, client_opts}, gen_server_opts)
   end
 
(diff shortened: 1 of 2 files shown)


lib/thrift/binary/framed/client.ex
```diff
@@ -81,14 +81,15 @@ defmodule Thrift.Binary.Framed.Client do
 
      - `send_timeout`: An integer that governs how long our connection waits when sending data.
 
-    `gen_server_opts`: A keyword list of options that control the gen_server behaviour.
+     - `gen_server_opts`: A keyword list of options that control the gen_server behaviour.
```
fishcakez: I think this might be easier to use with the GenServer options as top level. For tcp the list is not a `Keyword.t` so we can't have those options at the top level, and must nest them. For GenServer start I think we could support `:name`, `:spawn_opt` and `:debug`. The other one is`:timeout` and that is not helpful to us because we don't do anything in `init/1` so never block, and then we don't need to worry about a key clash.

Also these options have different meaning to when `:gen_server_opts` is used elsewhere, which is confusing, and breaks the abstraction I think.

lib/thrift/binary/framed/client.ex
```diff
@@ -81,14 +81,15 @@ defmodule Thrift.Binary.Framed.Client do
...
 
      - `timeout`:  The amount of time to wait (in milliseconds) for a reply from a `GenServer` call.
 
   """
   @spec start_link(String.t, (0..65_535), options) :: GenServer.on_start
```
fishcakez: While we do this change could we switch from `options` to `[option]`. It is no possible to compose a list type, but an element type can be composed with `[option | other_option_type]`.

lib/thrift/binary/framed/client.ex
thecodeboss: By top level, do you mean being able to specify the options like
```elixir
Client.start_link("localhost", 8080, name: __MODULE__)
```
instead of
```elixir
Client.start_link("localhost", 8080, gen_server_opts: [name: __MODULE__])
```
Also I've noticed that I misinterpreted that line to have a missing `-`, when actually it was intentionally unindented as a separate list of options from `tcp_opts`.

fishcakez: I did mean that yes but didn't notice this was just `-`. Lets go with this change :).

thecodeboss:
Made changes to how the options are passed based on the feedback, so they are now top-level. I think this makes the client a bit simpler to use when specifying options.

PR #328, opened 2018-03-04
Fix bug with enum numbering

thecodeboss:
There's a thrift definition in `test/fixtures/app/thrift/ThriftTest.thrift` called `Numberz` that looks like this:
```thrift
enum Numberz {
  ONE = 1,
  TWO,
  THREE,
  FIVE = 5,
  SIX,
  EIGHT = 8
}
```
Our previous solution was using `Enum.with_index` to assign indices, with special cases for those with special values. However, for this particular (malicious!) code, it resulted in
```elixir
...
  def(value_to_name(5)) do
    {:ok, :five}
  end
  def(value_to_name(5)) do
    {:ok, :six}
  end
...
```
and some other bad stuff showing up in our generated code.

### DIFF
--- a/lib/thrift/parser/models.ex
+++ b/lib/thrift/parser/models.ex
@@ -61,17 +61,15 @@ defmodule Thrift.Parser.Models do
 
     @spec new(charlist, %{charlist => enum_value}) :: t
     def new(name, values) do
-      values = values
-      |> Enum.with_index
-      |> Enum.map(fn
-        {{name, value}, _index} ->
-          {atomify(name), value}
-
-        {name, index} ->
-          {atomify(name), index + 1}
+      {_, values} = values
+      |> Enum.reduce({1, []}, fn
+        {name, value}, {_index, acc} ->
+          {value + 1, [{atomify(name), value} | acc]}
+        name, {index, acc} ->
+          {index + 1, [{atomify(name), index} | acc]}
       end)
 
-      %TEnum{name: atomify(name), values: values}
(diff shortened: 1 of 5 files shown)


lib/thrift/parser/models.ex
```diff
@@ -61,17 +61,14 @@ defmodule Thrift.Parser.Models do
...
     def new(name, values) do
-      values = values
-      |> Enum.with_index
-      |> Enum.map(fn
-        {{name, value}, _index} ->
-          {atomify(name), value}
-
-        {name, index} ->
-          {atomify(name), index + 1}
+      {_, values} = Enum.reduce(values, {1, []}, fn
```
jparise: Should the initial index start at 0 instead of 1?

Apache Thrift [initializes this value to -1](https://github.com/apache/thrift/blob/master/compiler/cpp/src/thrift/thrifty.yy#L488) and [uses a pre-increment](https://github.com/apache/thrift/blob/master/compiler/cpp/src/thrift/thrifty.yy#L528) to assign each implicit value. ([Explicit values are taken literally.](https://github.com/apache/thrift/blob/master/compiler/cpp/src/thrift/thrifty.yy#L517))

The result is that an enumeration with only implicit values starts at 0.

thecodeboss: You're absolutely right. I tested generating some enums and it indeed starts at 0. I hadn't checked, I was just replicating the functionality here that already started at 1.
Turns out this is a substantially bigger bug than I anticipated, all our enums had the potential to be off by one.

PR #329, opened 2018-03-19
Handle special floats like NaN, inf, -inf

thecodeboss:
Turns out Erlang/Elixir cannot use the binary pattern match `<<value::float-signed>> = binary` on `NaN`, `inf`, or `-inf`. It throws an argument error.

You can view [this Stack Overflow answer](https://stackoverflow.com/questions/38080582/why-is-infinity-not-allowed-in-erlangs-floats) for more details and a nice link to the IEEE specs explained in a readable way, and also [this commit in postgrex](https://github.com/elixir-ecto/postgrex/commit/0e6b6202d93718a8cc130b7d27bdfe9507bb3b00?diff=split) by @fishcakez where I got inspiration for the solution.

A few things to note:
* Denormalized floats can still be matched with `<<value::float-signed>>`, so I did not add special cases for them
* I've handled special values even in structures like maps, which should hopefully never have a `:NaN` as a key, but I think it's best that we support it and let clients handle any errors that result
* I imagine there's some small overhead caused by having to attempt a few extra matches on every float that is serialized/deserialized, and I don't know a good way to avoid it

Thanks to @pguillory for finding this; it was causing a crash in production, failing to deserialize some stuff.


test/thrift/generator/binary_protocol_test.exs
```diff
@@ -86,16 +86,27 @@ defmodule Thrift.Generator.BinaryProtocolTest do
...
+    assert_serializes %Double{val: 0.0},                       <<4, 0, 1, 0, 0, 0, 0, 0, 0, 0, 0, 0>>
+    assert_serializes %Double{val: 1.0},                       <<4, 0, 1, 1::signed-float, 0>>
+    assert_serializes %Double{val: 255.0},                     <<4, 0, 1, 255::signed-float, 0>>
+    assert_serializes %Double{val: :NaN},                      <<4, 0, 1, 0::1, 2047::11, 1::1, 0::51, 0>>
+    assert_serializes %Double{val: :inf},                      <<4, 0, 1, 0::1, 2047::11, 0::52, 0>>
+    assert_serializes %Double{val: :"-inf"},                   <<4, 0, 1, 1::1, 2047::11, 0::52, 0>>
+    assert_serializes %Double{val_map: %{}},                   <<13, 0, 2, 4, 4, 0, 0, 0, 0, 0>>
+    assert_serializes %Double{val_map: %{91.0 => 92.0}},       <<13, 0, 2, 4, 4, 0, 0, 0, 1, 91::signed-float, 92::signed-float, 0>>
+    assert_serializes %Double{val_map: %{:NaN => :inf}},       <<13, 0, 2, 4, 4, 0, 0, 0, 1, 0::1, 2047::11, 1::1, 0::51, 0::1, 2047::11, 0::52, 0>>
+    assert_serializes %Double{val_map: %{:NaN => :"-inf"}},    <<13, 0, 2, 4, 4, 0, 0, 0, 1, 0::1, 2047::11, 1::1, 0::51, 1::1, 2047::11, 0::52, 0>>
```
pguillory: Misses a couple of cases:
- `:inf` map key
- `:-inf` map key
- `:NaN` map value

pguillory: Sorry, kind of a nit, but found several errors previously through exhaustive testing.

lib/thrift/generator/struct_binary_protocol.ex
```diff
@@ -263,6 +272,15 @@ defmodule Thrift.Generator.StructBinaryProtocol do
...
   defp map_key_deserializer(:double, key_name, value_name, _file_group) do
     quote do
+      defp unquote(key_name)(<<0::1, 2047::11, 0::52, rest::binary>>, stack) do
+        unquote(value_name)(rest, :inf, stack)
+      end
+      defp unquote(key_name)(<<1::1, 2047::11, 0::52, rest::binary>>, stack) do
+        unquote(value_name)(rest, :"-inf", stack)
+      end
+      defp unquote(key_name)(<<_::1, 2047::11, _::52, rest::binary>>, stack) do
+        unquote(value_name)(rest, :NaN, stack)
```
fishcakez: We could have multiple NaNs disappear in this map as many possible keys are going to a single key, losing the values of all but one key. I think we should consider raising an error if can not define a one-to-one mapping.

Would we want to keep the sign and fraction information so that we can do loseless passthrough? That would mean %NaN{sign, fraction} or similar?

thecodeboss: I guess having a %NaN{sign, fraction} type makes sense. I'll edit this PR with such a type.

thecodeboss:
@fishcakez @jparise @pguillory bump

lib/thrift/nan.ex
```diff
@@ -0,0 +1,11 @@
+defmodule Thrift.NaN do
```
jparise: This could also just go in `thrift.ex`.

lib/thrift/protocol/binary.ex
```diff
@@ -39,15 +40,19 @@ defmodule Thrift.Protocol.Binary do
...
+  def serialize(:bool, false),     do: <<0::8-signed>>
+  def serialize(:bool, true),      do: <<1::8-signed>>
+  def serialize(:i8, value),       do: <<value::8-signed>>
+  def serialize(:i16, value),      do: <<value::16-signed>>
+  def serialize(:i32, value),      do: <<value::32-signed>>
+  def serialize(:i64, value),      do: <<value::64-signed>>
+  def serialize(:double, :inf),    do: <<0::1, 2047::11, 0::52>>
+  def serialize(:double, :"-inf"), do: <<1::1, 2047::11, 0::52>>
+  def serialize(:double, %NaN{sign: sign, fraction: frac}), do: <<sign::1, 2047::11, frac::52>>
+  def serialize(:double, :NaN),    do: <<0::1, 2047::11, 1::1, 0::51>>
```
jparise: Do we still need to match `:NaN` now that `%NaN{}` has been introduced?

lib/thrift.ex
```diff
@@ -29,4 +29,16 @@ defmodule Thrift do
...
     [:bool, :i8, :i16, :i32, :i64, :binary, :double, :byte, :string]
   end
+
+  defmodule NaN do
+    @moduledoc """
+    A struct for handling IEEE-754 NaN values.
+    """
+
+    @type t :: %NaN{sign: 0 | 1,
+                    fraction: (1..4_503_599_627_370_495)  # 2^52 - 1
```
jparise: Fraction can't be 0?

thecodeboss: If the fraction is 0, it is `:inf` or `:"-inf"`.

Issue #323, opened 2018-02-02
Remove dependency on `thrift` compiler and Erlang runtime


jparise:
- [x] Test serialization routines against static test data files
- [x] Test against our own client against our own server
- [x] Remove test case support for Thrift Erlang code generation
- [x] Remove Thrift Erlang runtime support files
- [x] Simplify our Travis CI test environment

pguillory:
+1. Is there a simple way we could still verify client/server interoperability with other languages? Could we integrate with the Apache test suite somehow, even if it's a manually run test procedure? Or mock out the other end of a TCP connection at the byte level based on the spec?

thecodeboss:
> Test serialization routines against static test data files
For this checkbox, I'm imagining we would have a script we can run offline to generate some static test data (or a mix task 😄), and this data gets checked into the repo for the tests to run against. The mix task could use the Erlang thrift implementation to generate the static test data.

This way, generating the static data would require the `thrift` compiler, but actually running the integration tests would not.

How does that sound? Thinking I could take a stab at this sometime soon.

jparise:
> For this checkbox, I'm imagining we would have a script we can run offline to generate some static test data (or a mix task 😄), and this data gets checked into the repo for the tests to run against. The mix task could use the Erlang thrift implementation to generate the static test data.

I actually don't think we need anything that advanced because I don't suspect we'll be regenerating it very often. For this level of testing, we just need to serialize some static "golden" data to disk and prove that we can (a) read them and (b) write data that looks identical to them. We could use the schema defined by [`ThriftTest.thrift`](https://github.com/pinterest/elixir-thrift/blob/master/test/fixtures/app/thrift/ThriftTest.thrift) to write a few large blobs, or we could have a directory of many small per-case files.

It might also be nice to write some [StreamData](https://github.com/whatyouhide/stream_data) generators, too, but that's a task for later.

The goal of this step is _just_ to test the correctness of the serialization routines themselves and not to be any kind of larger integration test.

thecodeboss:
Finished by #350

PR #330, opened 2018-03-30
fix test failed on windows

Xiaobin0860:


jparise:
@Xiaobin0860, could you remove the changes to the `test/support/src/*` files? Those are local copies of the Apache Thrift library. It looks like your changes were only for debug logging, and they don't appear to be related to the functionality you're fixing here.

Note that we're also planning for the removal of the Erlang bits in #323.

jparise:
Thanks, @Xiaobin0860!

Issue #301, opened 2017-12-20
Support CompactBinaryProtocol


dantswain:
Anyone thought about this?  I'm doing some stuff with [Parquet](http://parquet.apache.org/documentation/latest/) and that uses the compact protocol under the hood.

I'd have to dig deeper into the protocol, but from what I remember it's mostly the same as the binary protocol except some types are encoded differently, and some fields may not have fixed width.  My swagger idea for implementation would be to use the binary protocol code generation and make modifications on top of that as needed.

l may take a stab at this at some point if no one else claims it.

tompave:
Hi @dantswain, have you made any progress on this?

dantswain:
No, I haven't had time :(  I'm still very interested in it.

martyphee:
We're very interested in it also.  @dantswain do you have thoughts on how you would add this in?

dantswain:
The protocol is _mostly_ the same as the binary protocol, so my idea was to try to use the code generation that we already use for the binary protocol and make modifications as necessary to support the various differences.  It may need a two-pass sort of approach to handle the zig-zag encoding?  If you have existing code that can generate compact protocol binaries, that would be super helpful bc you can use that to generate test cases.

If you want more info on how the code generation works, I'd be glad to chat about it.

martyphee:
Thank you.  I might take you up on that.  Currently, Java is putting messages onto the bus so we have binary stream example to use/test with.  We're trying to get an Elixir project off the ground and this is one stumbling block.

paulanthonywilson:
Hey, so I've made a start at looking at how this might be done. It's nowhere near ready for a PR here, but I've made this to  make it easier to follow the approach and get feedback. https://github.com/CultivateHQ/elixir-thrift/pull/1 . 

(I've got as far as serialising the test `Bool` struct. https://github.com/CultivateHQ/elixir-thrift/blob/compact-protocol/test/thrift/generator/compact_protocol_test.exs#L57-L82)

paulanthonywilson:
I'm not entirely sure about translating this test / behaviour to the compact protocol:

https://github.com/pinterest/elixir-thrift/blob/50c575831e1c37af010b0ff19f0b42c6e1a2d925/test/thrift/generator/binary_protocol_test.exs#L119

The binary version wraps i16 overflows on encoding, so 32768 becomes -32768 and 65536 is 0. 

Should the compact protocol do the same, even though it's encoded in a zigzag varint?

[...]

PR #333, opened 2018-04-18
Compact protocol serialisation and deserialisation of Structs

paulanthonywilson:
This partially addresses https://github.com/pinterest/elixir-thrift/issues/301, implementing compact protocol for Structs.

It's a bit of a monster, I'm afraid. There was quite a lot to do.  From `Thrift.Generator.StructCompactProtocol`

```
  This is based somewhat on `Thrift.Generator.StructBinaryProtocol`, but implemented
  for the compact protocol.

  https://github.com/apache/thrift/blob/master/doc/specs/thrift-compact-protocol.md has been
  used as the description of the binary protocol, verified by testing with the Thrift
  Ruby and Python implementations. There are a few discrepancies between the document and
  the discovered implementation; where these were found the discovered implementation
  was used. These are:

  * Doubles appear to be actually encodes as little endian
  * The document states different codes to indicate types contained in maps, lists, and sets to
  those for Structs; in fact the same codes are used and 1 is used to indicate a Bool.
  * For element boolean types, false is encoded as 2 rather than 0

  The compact protocol introduces some complications which necessitate a divergence from
  the approach taken for the binary protocol.

  Field headers may be of two forms, depending on whether the type and field id information
  can fit into a single byte; this means that two deserialisation matches must be created for each
  field.

  In the short form field header, the field id is not used directly; rather the difference
  between the previously encoded field id and the field id is used, termed the _delta_. This means
  that we must maintain information about previous fields while serialising and deserialising.

  List and Set header information may also be in two forms, depending on whether the size and
  element-type information can fit into a single byte, complicating deserialisation.

  Use of varint encoding for size in various places reduces the scope for using binary matching in
  many places.
```

Other things to consider are:

* Only Struct serialisation is supported. I haven't looked at messages or the RPC part yet. This part has taken a while and I thought I should try and get some feedback.
* Right now, each generated Struct module has both _Compact_ and _Binary_ serialisers generated. I suspect it would be better to only generate the compact form if (say) a configuration flag is set.


fishcakez:
Thanks for looking into this @paulanthonywilson, it definitely is a big PR so may take a little time to get feedback on the whole :)

paulanthonywilson:
Yeah, totally @fishcakez 

Sorry about the size it kind of seems a bit all or nothing :-/

paulanthonywilson:
Hey, so ... 

I was thinking about this monster PR and how it's a monstrous amount of code to add to this codebase. I was thinking about how it could be smaller.

What Would you think about making the protocol configurable? So the binary protocol is the default standard, and remains in this repo but an alternative protocol like compact (or JSON) could be plugged in?

jparise:
Sorry for the delay on this, @paulanthonywilson. The team has been quite busy lately, and I'm out of the office on paternity leave, so my time has been _extra_ limited. 🍼

> What would you think about making the protocol configurable? So the binary protocol is the default standard, and remains in this repo but an alternative protocol like compact (or JSON) could be plugged in?

My initial reaction to that idea is -0. If these were "non-standard" components of the Thrift protocol suite, then that would make sense, but I think there's value in everything being tested, packaged, and maintained together in a single library.

... but if you have a specific idea on how to handle the separation, I'm definitely open to considering it. If you want to pursue that, please open an "RFC" issue, and we can discuss it there.

fishcakez:
Hi @paulanthonywilson! I will try to get through this PR over the next few weeks. As we have taken a while to review this I may make some minor changes to the branch in order to get it merged. Likely there will be follow up issues to improve protocol selection for code generation.

paulanthonywilson:
Hey @fishcakez, that's great - thanks. Do bear in mind that's it's not currently complete - it only covers encoding / decoding structs as I wanted to get feedback before making the PR even bigger. (Not a passive-agressive complaint - there's no urgency about this on my part.)

I meant to reply to @jparise to see congrats on 🍼; also sorry for not getting round to putting the RFC issue in.

paulanthonywilson:
@fishcakez Ok, that's remerged \o/

(I'm currently on family holiday with dodgy Internet, so I may be slow in replying to anything over the next week.)

fishcakez:
Thanks @paulanthonywilson, enjoy your holiday, lets revisit when you're back :)

paulanthonywilson:
Oh, wow. I just noticed this was still here. So ...  here's a story. 

About 3 years ago I ran an agency that primarily did Ruby work. But I wanted us to be an agency that primarily did Elixir work. One of our major clients senior developers was interested in switching to Elixir but said that they were moving to Compact Thrift for message serialisation between services, in different languages, and it would be cool if this library would support that. And I was like "how hard could it be....".

So I got this far, which supported all that was really needed, except that they changed their strategy to using Protobuffs. (And they decided to standardise on Go instead of Elixir.)

Anyway ... it's been a while (thought it seems longer than 3 years) and I don't remember a whole lot about the PR except that I did have a fun writing it. But it's time to move on.

pour_out

PR #341, opened 2018-05-11
Add option for allowing both SSL and plain connections on same port

thecodeboss:
By inspecting the first byte of traffic on the socket, we can determine whether the connection is a plain connection or an SSL connection. By having a single server that handles both connection types, it simplifies operations so we do not need to run two servers on different ports.

`:gen_tcp` makes this difficult as there is no `peek` option for sockets like there are in C, but instead we can recv the first bytes and then `unrecv` them back to the socket. Before doing this though, we ensure that `packet: 4` has not been set.

I added a bunch more tests so we cover a plain client trying to connect to a server that requires SSL, and also the optional cases.

### DIFF
--- a/lib/thrift/binary/framed/protocol_handler.ex
+++ b/lib/thrift/binary/framed/protocol_handler.ex
@@ -30,16 +30,30 @@ defmodule Thrift.Binary.Framed.ProtocolHandler do
 
     {recv_timeout, tcp_opts} = Keyword.pop(tcp_opts, :recv_timeout, @default_timeout)
 
-    transport_options = Keyword.put(tcp_opts, :packet, 4)
-    transport.setopts(socket, transport_options)
+    with {:ok, first_bytes} <- :gen_tcp.recv(socket, 4, recv_timeout),
+         :ok <- :gen_tcp.unrecv(socket, first_bytes) do
+      <<first_byte :: 8-unsigned, _ :: binary>> = first_bytes
 
-    maybe_ssl_handshake(socket, ssl_opts, server_module, handler_module, recv_timeout)
+      transport_options = Keyword.put(tcp_opts, :packet, 4)
+      transport.setopts(socket, transport_options)
+
+      maybe_ssl_handshake(socket, first_byte, ssl_opts, server_module, handler_module, recv_timeout)
+    else
+      {:error, closed} when closed in [:closed, :econnreset, :timeout] ->
+        :ok = transport.close(socket)
+
+      {:error, reason} ->
+        # :ssl.format_error handles posix errors as well as ssl errors
+        Logger.info(fn -> "#{inspect handler_module} (#{inspect self()}) connection error: #{:ssl.format_error(reason)} (#{inspect reason})" end)
+        :ok = transport.close(socket)
+    end
   end
 
-  defp maybe_ssl_handshake(socket, ssl_opts, server_module, handler_module, timeout) do
+  defp maybe_ssl_handshake(socket, first_byte, ssl_opts, server_module, handler_module, timeout) do
     with {:ok, ssl_opts} <- SSL.configuration(ssl_opts),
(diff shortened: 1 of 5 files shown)


lib/thrift/binary/framed/protocol_handler.ex
```diff
@@ -50,6 +64,23 @@ defmodule Thrift.Binary.Framed.ProtocolHandler do
     end
   end
 
+  defp maybe_ssl_accept(socket, 0x16, _optional, ssl_opts, timeout) do
+    with {:ok, ssl_sock} <- :ssl.ssl_accept(socket, ssl_opts, timeout) do
```
jparise: `case` might be simpler here, although I don't feel strongly.

test/thrift/transport/ssl_test.exs
```diff
@@ -9,4 +9,11 @@ defmodule Thrift.Transport.SSLTest do
       assert {:error, error} == SSL.configuration([enabled: true, configure: fn -> {:error, error} end])
     end
   end
+
+  describe "optional?/1" do
+    test "does the right thing" do
+      assert {true, [certfile: "/some/path"]} = SSL.optional?([certfile: "/some/path", optional: true])
```
jparise: Can use `==` because we're not assigning the match.

lib/thrift/binary/framed/protocol_handler.ex
```diff
@@ -50,6 +64,23 @@ defmodule Thrift.Binary.Framed.ProtocolHandler do
     end
   end
 
+  defp maybe_ssl_accept(socket, 0x16, _optional, ssl_opts, timeout) do
```
jparise: Would it be useful to define `@ssl_magic_byte 0x16` or similar to self-document this value?

lib/thrift/transport/ssl.ex
```diff
@@ -30,6 +31,11 @@ defmodule Thrift.Transport.SSL do
     end
   end
 
+  @spec optional?([option]) :: {boolean, [:ssl.ssloption]} | {:error, Exception.t}
+  def optional?(opts) do
+    Keyword.pop(opts, :optional, false)
+  end
```
fishcakez: I think it would simplify the API if we merged `configuration` and `optional` so that `configuration` return `{:require | :optional, opts} | ..}`.

lib/thrift/binary/framed/protocol_handler.ex
```diff
@@ -30,16 +30,30 @@ defmodule Thrift.Binary.Framed.ProtocolHandler do
 
     {recv_timeout, tcp_opts} = Keyword.pop(tcp_opts, :recv_timeout, @default_timeout)
 
-    transport_options = Keyword.put(tcp_opts, :packet, 4)
-    transport.setopts(socket, transport_options)
+    with {:ok, first_bytes} <- :gen_tcp.recv(socket, 4, recv_timeout),
```
fishcakez: We receive 4 bytes but check only 1. Does it break if we only receive 1?

thecodeboss: Yes it breaks if we only receive 1. And it took me hours to debug why it wouldn't work.

lib/thrift/transport/ssl.ex
thecodeboss: That's a great idea, I'll do that

nichochar:
I am a little concerned that this dual server opens up vulnerabilities (someone spoofing the first byte and pretending they don't want an SSL handshake). 
I think that given that we're using this for convenience of the migration, this is fine, but I'm not sure it is secure to be used in prod as a server. 

Maybe we could add a comment in the code disclaiming this?

thecodeboss:
@nichochar I don't think there's a security vulnerability in this case. If they spoof the first byte, they either:
a) fail the SSL handshake because they can't spoof that part, or
b) simply get downgraded to a plain connection. But the person who set this to optional explicitly allowed that to happen by setting the `optional: true` flag.

PR #347, opened 2018-06-13
Small improvements to Schema.new/3

jparise:
- build the schema using a single `merge/2` reduction
- `path` can be nil when we're parsed from an in-memory string
- remove the `model` type that isn't doing much for us
- remove the unused `thrift_namespace` field


jparise:
Construction may have needed two passes (headers and then definitions) previously, perhaps because the second pass relied on values generated by the first pass. We don't have any of those dependencies now, and the order is still consistent, so I believe that part of this change is sane.

PR #348, opened 2018-06-13
Add a .formatter.exs configuration file

jparise:
The only change from the default configuration is the inclusion of the
`bench/` directory to the list of inputs.

Also, apply the formatter to the `bench/*.exs` source files.

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -1,0 +1,3 @@
+[
+  inputs: ["{mix,.formatter}.exs", "{bench,config,lib,test}/**/*.{ex,exs}"]
+]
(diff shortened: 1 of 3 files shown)


jparise:
Alternatively, we could drop the current benchmark scripts entirely because they're based on the Apache Thrift Erlang library, which we're attempting to move away from. What do you think, @fishcakez?

fishcakez:
We can drop the benchmarks after #350

jparise:
Yes, that’s while I’ve been holding this. I’ll work up a new PR that just removes them.

jparise:
Dropping this in favor of #351.

PR #349, opened 2018-06-15
Quality of life improvements

scohen:
Just a couple quality of life improvements. 

We needed to increase the max connections beyond Ranch's 1024, which proved impossible because you can't specify transport_opts easily.

Also fixed a 1.6 compile due to the ordering of catch and rescue.

### DIFF
--- a/lib/thrift/generator/binary/framed/server.ex
+++ b/lib/thrift/generator/binary/framed/server.ex
@@ -99,8 +99,6 @@ defmodule Thrift.Generator.Binary.Framed.Server do
     quote do
       try do
         unquote(quoted_handler)
-      rescue
-        unquote(rescue_blocks)
       catch kind, reason ->
         formatted_exception = Exception.format(kind, reason, System.stacktrace)
         Logger.error("Exception not defined in thrift spec was thrown: #{formatted_exception}")
@@ -108,6 +106,8 @@ defmodule Thrift.Generator.Binary.Framed.Server do
           type: :internal_error,
           message: "Server error: #{formatted_exception}")
         {:server_error, error}
+      rescue
+        unquote(rescue_blocks)
       end
     end
   end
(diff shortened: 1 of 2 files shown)


lib/thrift/generator/binary/framed/server.ex
```diff
@@ -99,15 +99,15 @@ defmodule Thrift.Generator.Binary.Framed.Server do
...
-      rescue
-        unquote(rescue_blocks)
       catch kind, reason ->
         formatted_exception = Exception.format(kind, reason, System.stacktrace)
         Logger.error("Exception not defined in thrift spec was thrown: #{formatted_exception}")
         error = Thrift.TApplicationException.exception(
           type: :internal_error,
           message: "Server error: #{formatted_exception}")
         {:server_error, error}
+      rescue
```
nichochar: Out of curiosity, should `rescue` always come after `catch` ? If so maybe we can add some static analysis and catch this at the erbert level

thecodeboss: Elixir 1.6 has a warning built in for this I believe

scohen: @nichochar, yes, rescue should always come after catch (at least according to the new style rules). I don't think there's any technical reason for this, but 1.6 now emits a warning when this isn't the case.

scohen:
Does anyone know why the test failed?

jparise:
@scohen, I don't see the original test failure, but it looks green now. Did you trigger a rebuild?

There are still some flaky service tests in the suite due to the way that the Erlang Thrift server non-deterministically closes its sockets and shuts down its processes.

scohen:
That's... odd... I see it as green as well. 
Going to merge.

scohen:
The tests that failed were SSL related, so it might have been the erlang quirkiness you mentioned.

PR #350, opened 2018-07-21
Remove dependency on 'thrift' compiler and Erlang runtime

thecodeboss:
Completes #323


test/thrift/protocol/binary_test.exs
```diff
@@ -15,6 +15,26 @@ defmodule BinaryProtocolTest do
     :erlang.apply(deserializer_mod, deserializer_fn, [serialized])
   end
 
+  def serialize(module, struct) do
```
jparise: Minor, but these could be `defp`.

thecodeboss:
> Should we include some kind of mechanism to regenerate the .thriftbin reference files?

You had mentioned in #323 that you don't think it's necessary to do that, as we shouldn't really need to regenerate the reference files. I think that holds. I think writing a mechanism would be a decent amount of work, and I don't think we'd get too much value from it.

jparise:
> You had mentioned in #323 that you don't think it's necessary to do that, as we shouldn't really need to regenerate the reference files. I think that holds. I think writing a mechanism would be a decent amount of work, and I don't think we'd get too much value from it.

The case that comes to mind (which I hadn't considered before) is for the higher-level types where our current serialized data isn't exhaustive of their capabilities. For example, if we want to test a new `enum` case, we'd have to modify them.

I agree that it will happen rarely. Maybe a good compromise is just adding a brief `README.md` to the `test/static/` directory (which maybe should be named `test/data/`?) explaining the generation steps manually.

fishcakez:
I think its important to have instructions to generate test data because its simplifies the process for someone to write a test verifying a bug fix or new feature. I don't think that having a sophisticated mix task to generate thrift data from "source" is required. We already have a use case for instructions as someone will need to repeat the process for compact protocol: https://github.com/pinterest/elixir-thrift/pull/333.

My initial thought is that thriftpy scripts would be the simplest way to generate the data in future because it creates the thrift modules at import time and data structures/file could be passed as command line arguments. As this will be run manually and rarely we don't need to be concerned with requiring python.

thecodeboss:
I added a Python script for generating the test data, roughly equivalent to what I used initially to generate but just a bit more user friendly.
I also noticed that I got non-deterministic results when re-generating test data, as there is no order guarantee for set/map serialization. I adjusted the `assert_serde` method accordingly to not directly compare the serialized data as it can differ.

test/data/generate.py
```diff
@@ -0,0 +1,67 @@
...
+
+See `README.md` for instructions on running this script.
+"""
+from glob import iglob
+import os
+import re
+import shutil
+import subprocess
+import sys
+import test_data
```
jparise: `from . import test_data`

test/data/generate.py
```diff
@@ -0,0 +1,67 @@
...
+    try:
+        os.mkdir(dir)
+    except FileExistsError:
+        pass
+
+
+def find_thrift_definitions():
+    thrift_definitions = []
+    for filename in iglob('../**/*.exs', recursive=True):
+        file_contents = ''
```
jparise: You don't need to declare `file_contents` before the `with` block.

test/data/generate.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+def write_thrift_files(thrift_definitions):
+    mkdir('thrift')
+    thrift_files = []
+    for filename, definition in thrift_definitions:
+        full_path = os.path.join('thrift', filename)
+        with open(full_path, 'w') as f:
+            f.write(definition)
+            thrift_files.append(full_path)
```
jparise: This line doesn't need to be within the `with` block.

test/data/generate.py
```diff
@@ -0,0 +1,67 @@
...
+    return thrift_files
+
+
+def generate_python_files(thrift_files):
+    mkdir('generated')
+    for thrift_file in thrift_files:
+        command = 'thrift -I thrift --gen py --out generated'.split(' ') + [thrift_file]
```
jparise: You can just use `.split()`; the default `sep` argument behavior is already what you want.

test/data/generate.py
thecodeboss: Relative imports like this one are only allowed in a package. See [this Stack overflow answer](https://stackoverflow.com/questions/16981921/relative-imports-in-python-3).
It wouldn't make sense to turn this script into a package since it is supposed to be a standalone thing.

PR #352, opened 2018-07-22
Run mix format

thecodeboss:
I ran `mix format` over the codebase, and made some minor tweaks to readability where appropriate.

This is somewhat a RFC as I am unsure if we want this or not.


fishcakez:
I am happy for using formatter, and if we are going to use it can we add a check in travis to make sure code is correctly formatted in future?

jparise:
Something looks wrong about the way that Travis is expanding the testing matrix.

Maybe the easier thing to do is to consolidate this check and the coveralls check under a single environment variable. I assume we'll also want to run those code quality checks together and only in a single environment.

thecodeboss:
@jparise yeah, I was trying to debug what was going on, and I really don't have an answer. I'll try one environment variable

.travis.yml
```diff
@@ -14,13 +14,10 @@ matrix:
       otp_release: 19.3
     - elixir: 1.7
       otp_release: 21.0
-      env:
-        - CHECK_FORMAT=true
-        - COVERALLS=true
+      env: EXTRA_CI_CHECKS=true
```
jparise: Just `EXTRA_CHECKS`? The `CI` is a little redundant in this context. 😉

ci/tests.sh
```diff
@@ -2,18 +2,19 @@
...
 #
-# This script could be used for local testing as long as COVERALLS is not set.
+# This script could be used for local testing as long as EXTRA_CI_CHECKS is not set.
 
 export MIX_ENV=test
 
-if [ "$COVERALLS" = true  ]
+if [ "$EXTRA_CI_CHECKS" = true  ]
 then
+  echo "Checking that all code is formatted"
```
jparise: Looking at Travis output, I think this would be a little easier to read in the log:

```
Checking code format (`mix format --check-formatted`) ...
```

PR #355, opened 2018-07-29
Fix flakey tests

fishcakez:


test/thrift/generator/service_test.exs
```diff
@@ -371,8 +371,12 @@ defmodule Thrift.Generator.ServiceTest do
 
   thrift_test "clients exit if they try to use a closed client", ctx do
     Process.flag(:trap_exit, true)
+    # Wait for client to connect
+    client = ctx.client
```
jparise: Or match `%{client: client}` above instead of using `ctx`.

Issue #288, opened 2017-10-03
RFC - Documentation needs


dantswain:
I'm starting to take a pre-2.0 pass through the docs.  Some things will be easy enough to just knock out, some things I think could use some discussion.

* Contributing guidelines - @jparise does Pinterest have any standard open source contributing guidelines?
* Which modules should actually be `@moduledoc`d? - Some of the code is not really intended for public API use, so `@moduledoc false` may be more appropriate.
* Examples - I find it really useful when a project like this has a simple example project to reference - either in a subdirectory or a separate repository.  Is that something we want?  Is it something we want to do before 2.0?
* What else do we need to document?

jparise:
I'll own the Contributor Guidelines topic separately. It's something we've been meaning to standardize across all of Pinterest's open source projects.

I agree with a general `@moduledoc` audit. We probably only need to document the top-level modules like `Thrift.Parser` and `Thrift.Generator`, and not all of the nested modules within them.

And I'm +1 on including an example project.

fishcakez:
ping @jon @dantswain

PR #362, opened 2018-07-29
Add an initial contributing guide

jparise:
This will definitely expand and evolve, but I think it's a good start.

See #288

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,8 +1,10 @@
 # Contributing
 
-First off, thanks for taking the time to contribute! We welcome contributions
-from everyone, and this guide will answer some common questions about how this
-project works.
+First off, thanks for taking the time to contribute! This guide will answer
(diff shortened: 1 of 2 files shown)


CONTRIBUTING.md
```diff
@@ -0,0 +1,43 @@
...
+First off, thanks for taking the time to contribute! We welcome contributions
+from everyone, and this guide will answer some common questions about how this
+project works.
+
+## Making Changes
+
+1. Fork this repository to your own account
+2. Make your changes and verify that `mix test` passes
+3. Commit your work and push to a topic branch on your fork
+4. Submit a [pull request](https://github.com/pinterest/elixir-thrift/compare/)
```
fishcakez: Should we mention what to expect once a PR is open? I think this is the most likely stage where expectations can diverge. Something like:
5. Participate in mutual review process hopefully leading to merging the contribution.

PR #364, opened 2018-07-29
Remove the Travis and Coveralls status badges

jparise:
It doesn't seem useful to include these "live" icons as part of our
published documentation, especially for tagged versions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,8 +1,5 @@
 # A Pure Elixir Thrift Library
 
-[![Build Status](https://travis-ci.org/pinterest/elixir-thrift.svg?branch=master)](https://travis-ci.org/pinterest/elixir-thrift)
-[![Coverage Status](https://coveralls.io/repos/pinterest/elixir-thrift/badge.svg?branch=master&service=github)](https://coveralls.io/github/pinterest/elixir-thrift?branch=master)
-
 This package contains an implementation of [Thrift](https://thrift.apache.org/) for Elixir.
 It includes a Thrift IDL parser, a code generator, a binary framed client and a binary framed server.
 


thecodeboss:
> The GitHub UI already shows the project's build status

Where does it show this? I only see the badges on the [elixir-thrift project page](https://github.com/pinterest/elixir-thrift).

jparise:
> Where does it show this? I only see the badges on the elixir-thrift project page.

It shows up as a green check mark (or red x, for failure) next to the project name:

screen shot 2018-07-29 at 6 46 36 pm

jparise:
Actually, I just realized that's because of the [Refined GitHub extension](https://github.com/sindresorhus/refined-github) I'm using.

thecodeboss:
I think we should find a solution where we keep the build status visible in the readme, but don't publish it in our docs. For example, the [core Elixir project page](https://github.com/elixir-lang/elixir) has a badge for the current build status, but obviously this doesn't show up in the docs.

One option is to have a readme that does not get published with our docs, and only shows up on GitHub. Then our docs could use the `Thrift` module as the homepage, which would be much more detailed than the readme. This is how [ecto](https://github.com/elixir-ecto/ecto) does it ([here are the docs](https://hexdocs.pm/ecto/Ecto.html)).

thecodeboss:
> Actually, I just realized that's because of the Refined GitHub extension I'm using.

Also thanks for mentioning, I didn't know about this. Seems like a really awesome extension!

jparise:
> One option is to have a readme that does not get published with our docs, and only shows up on GitHub. Then our docs could use the Thrift module as the homepage, which would be much more detailed than the readme.

I think this would be an overall improvement so I'm going to pursue it. Thanks for the suggestion!

Issue #343, opened 2018-05-29
Split generated serde code from transport handling by introducing middlewares


fishcakez:
Currently we generate the client and server code for TFramedTransport and there is somewhat tight coupling. However there are four different parts to this end to end pipeline:

* Client, which generates a set of functions representing the service
* Protocol serde
* Transport
* Server behavior, which generates the callbacks required to act as the service

To decouple, we provide a common data structure that is passed through the pipeline on each of the peers:
```elixir
%Thrift.Message{name: String.t, type: :call | :oneway | :reply | :exception, seq_id: integer, payload: struct()}
```

On the client side the function for a call (or oneway) would create a `%Thrift.Message{}` with the method name, request type and payload as the request struct. This would be passed to the protocol layer, which replaces the request struct in the message with a new struct (e.g. `%Thrift.Binary{}`) containing the serialized request struct. The protocol then passes the message to the transport layer (e.g. `Thrift.Binary.Framed.Client`) which converts the message to data and sends it on the socket. The transport layer then receives the response, deserializes into a new `Thrift.Message` with serialized payload struct (e.g. `%Thrift.Binary{}`) and returns the message to the protocol layer. The protocol layer deseralizes the payload to the response struct, puts that in the message's payload and returns the message to the generated client function. Finally the client function handles the response struct and converts to the function's result,
i.e. `{:ok, result} | {:error, Exception.t}`.

This means our client side control flow is `Client -> Protocol -> Transport -> Protocol -> Client`. For the server side we reserve this to `Transport -> Protocol -> Server -> Protocol -> Transport`.

On the server side the transport layer (e.g. `Thrift.Binary.Framed.Server`) receives data from the socket, deserializes this into a message containing the method name, request type, sequence id and serialized payload (e.g. `Thrift.Binary` struct). The message is passed to the protocol layer, which deserializes the payload to the request struct and passes the message to the server layer. The server layer unrolls the request struct and dispatches it to the server's callback module, which handles the request and returns a response. The server layer turns this response into the response struct, puts it in the message struct with the new type (e.g. `:reply`) and returns it to the protocol layer. The protocol layer serializes the response struct into its own payload (e.g. `%Thrift.Binary{}`) and returns it to the transport layer. The transport layer serializes the message and sends it over the socket.

Lets take this through an example with the following schema:
```thrift
namespace elixir Example

service Service
{
  string ping(),
}
```
The generated client will generate code approximately like:
```elixir
defmodule Example.Service.Client do
[...]

PR #365, opened 2018-07-30
Remove stale TODO.md file

thecodeboss:
This file is very out-of-date, and there isn't much reason to use it given we have the [Issues page](https://github.com/pinterest/elixir-thrift/issues), which provides much more detailed tagging and tracking on these items.


jparise:
Anything in here that should be captured in an issue before we delete it?

fishcakez:
> Anything in here that should be captured in an issue before we delete it?

We still want a header based protocol or transport, likely TTwitter or THeader.

thecodeboss:
> We still want a header based protocol or transport, likely TTwitter or THeader.

I created #369, which has some overlap with #343, but it's nice to break it out into its own issue.
Looking through this TODO list, I don't see any others I think are worth filing issues for, so I'll close this now.

PR #374, opened 2018-08-01
Use the Thrift module as the main docs page

jparise:
This adds some initial high-level documentation to the `Thrift` module.
There are still many details that only exist in README.md, but this is
enough for us to remove that file from the generated documentation.

This change also adds a section grouping for the Thrift.AST modules.

See #288

### DIFF
--- a/lib/thrift.ex
+++ b/lib/thrift.ex
@@ -1,12 +1,55 @@
 defmodule Thrift do
   @moduledoc ~S"""
-  Thrift provides [Apache Thrift](http://thrift.apache.org/) support:
+  [Thrift](https://thrift.apache.org/) implementation for Elixir including a
+  Thrift IDL parser, a code generator, and an RPC system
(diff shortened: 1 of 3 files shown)


lib/thrift.ex
```diff
@@ -1,12 +1,55 @@
...
+
+      compilers: [:thrift | Mix.compilers]
+
+  It's important to add `:thrift` *before* the `:elixir` compiler entry. The
+  Thrift compiler generates Elixir source files, which are in turn compiled by
+  the `:elixir` compiler.
+
+  Configure the compiler using a keyword list under the top-level `:thrift`
+  key. The only required compiler option is `:files`, which defines the list
+  of Thrift files to compile.
```
thecodeboss: Maybe add a note here with, "see `Mix.Tasks.Compile.Thrift` for the other available options".

PR #376, opened 2018-08-06
Revise Travis CI configuration

jparise:
- Treat compilation warnings as errors
- Report coveralls for all build environments
- Conditionally run `mix format` from inside `.travis.yml`
- Cache the `deps/` directory instead of mix archives
- Only include credo in the :dev environment
- Only include excoveralls in the :test environment


.travis.yml
```diff
@@ -14,15 +14,19 @@ matrix:
...
+  - mix deps.get --only test
+
+before_script:
+  - MIX_ENV=test mix compile --warnings-as-errors
 
 script:
-  - MIX_ENV=test mix deps.compile
-  - ./ci/tests.sh
+  - if [[ "$TRAVIS_ELIXIR_VERSION" == "1.7"* ]]; then mix format --check-formatted; fi
+  - MIX_ENV=test mix coveralls.travis
```
fishcakez: Whats the consequence of reporting for all environments?

jparise: Coveralls.io now appears to handle this well. It lists individual coverage values for each build environment. Previously each environment would race to report, and either the first or last would win (I don’t recall).

PR #379, opened 2018-08-06
Expand the code generation documentation

jparise:
This explains how Thrift definitions are mapped to generated modules as
well as how namespaces are specified.

### DIFF
--- a/lib/thrift.ex
+++ b/lib/thrift.ex
@@ -90,7 +90,10 @@ defmodule Thrift do
 
   ### Namespaces
 
-  Thrift uses the `namespace` keyword to define a language's namespace:
+  The generated modules' namespace is determined by the `:namespace` compiler
+  option, which defaults to `Thrift.Generated`. Individual `.thrift` files can
+  specify their own namespace using the `namespace` keyword, taking precedence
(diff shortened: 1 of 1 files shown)


lib/thrift.ex
```diff
@@ -51,6 +51,58 @@ defmodule Thrift do
...
+      struct User {
+        1: i64 id,
+        2: string username,
+      }
+
+      service UserService {
+        bool ping(),
+        User get(1: i64 id) throws (1: UserNotFound e),
+        bool delete(1: i64 id),
+      }
```
thecodeboss: Maybe we should put this code in backticks so we can specify the language
````
```thrift
# thrift code
```
````
Hex docs use highlight.js for highlighting, which supports thrift, so in theory it will render nicely. In practice I'm not sure if it will work.

jparise: It unfortunately doesn't, perhaps due to ex_doc's recent switch to [makeup](https://hex.pm/packages/makeup).

... but I'll change the Thrift examples to fenced code blocks because they look nicer unhighlighted than they do highlighted as Elixir code, and maybe we'll get Thrift highlighting support in the future.

PR #380, opened 2018-08-06
Upgrade to ranch 1.6

jparise:
This is a minor upgrade for us because we don't use many of Ranch's
features. It's useful to upgrade now to avoid any new code being added
that might use recently deprecated features that will be removed in
Ranch 2.0.

https://ninenines.eu/docs/en/ranch/1.6/guide/migrating_from_1.5/

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -79,7 +79,7 @@ defmodule Thrift.Mixfile do
 
       # Runtime
       {:connection, "~> 1.0"},
-      {:ranch, "~> 1.5"}
+      {:ranch, "~> 1.6"}
     ]
   end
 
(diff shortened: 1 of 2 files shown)


thecodeboss:
Perhaps we should integrate with [Dependabot](https://dependabot.com/) to automate these dependency bumps. It's great because it submits PRs whenever dependencies can be bumped. I use it on one of my own projects and it works great.

thecodeboss:
> That seems like it could be useful. How do you handle the cases where project code should also be updated?

It allows maintainers to edit the PRs it posts, so if there are project changes required someone will need to make edits. Most of the time I've found the merges are pretty clean and don't require manual intervention.

It also supports commands like `@dependabot ignore this [patch|minor|major] version` and `@dependabot ignore this dependency` if you don't want to update a dependency. Useful in cases where upgrading the version is not trivial.

Here's an example of a PR it submitted for me to upgrade ranch from 1.5 to 1.6: https://github.com/thecodeboss/minecraft/pull/4

jparise:
> Do newer versions of ranch support gracefully shutdown?

I'm not sure, and I don't see it listed in the [2.0 roadmap](https://ninenines.eu/docs/en/ranch/1.6/guide/upcoming_2.0_changes/). Our use of Ranch is pretty limited to just the client, and we could probably drop it if we implemented our own connection pool.

PR #381, opened 2018-08-06
Revise the README to reflect version 2.0 status

jparise:
The updated Getting Started section is now a reduced version of the
previous text because many of those details are now in the formal module
documentation.

This change also adds Hex version and docs badges.


README.md
```diff
@@ -10,121 +12,42 @@ The serialization and deserialization code that is generated by this project is
...
+Elixir.
 
-```elixir
-{:thrift, "~> 2.0"}
-```
+## Getting Started
 
-Or to track the GitHub master branch:
+Until version 2.0 is released, you'll need to track the master branch
+directly:
```
thecodeboss: Should we release a `2.0.0-rc1` so folks can use Hex instead of GitHub directly? I see our mix file has `2.0.0-dev`, but having `rc` releases would give more flexibility for releasing more than one development version.

jparise: Yes, I think we're getting pretty close to that point. I'd like to get some more of the documentation updated before we invite the first round of feedback.

PR #383, opened 2018-08-07
Fully qualify the Thrift.Parser types

jparise:
This fixes both ex_doc's rendering and dialyzer's type resolution.

### DIFF
--- a/lib/thrift/ast.ex
+++ b/lib/thrift/ast.ex
@@ -19,7 +19,7 @@ defmodule Thrift.AST do
     namespaces, each with its own language scope.
     """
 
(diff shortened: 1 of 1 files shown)


jparise:
> Looks good, I'm curious what motivated the change. Dialyzer warnings or something? Or just looking for quality improvements?

I updated the summary above, but I noticed that ex_doc wasn't rendering the types correctly, and then realized that dialyzer wasn't resolving `line()` either. I may have broken that when I factored out the `Thrift.AST` module.

We should consider running dialyzer in CI once we have a clean build. It looks like we have a few other warnings right now.

thecodeboss:
> We should consider running dialyzer in CI once we have a clean build. It looks like we have a few other warnings right now.

Definitely.

PR #385, opened 2018-08-10
Document the AST header types (Include and Namespace)

jparise:
Also remove the unused `Thrift.AST.all()` type and make some small
improvements to the `Thrift.Parser` docs.

### DIFF
--- a/lib/thrift/ast.ex
+++ b/lib/thrift/ast.ex
@@ -2,7 +2,7 @@ defmodule Thrift.AST do
   @moduledoc """
   Thrift Abstract Syntax Tree
 
-  `Thrift.Parser` returns as a tree of these structures, starting with a
+  `Thrift.Parser` returns a tree of these structures, starting with a
   `Thrift.AST.Schema` node at the root.
 
   ## Headers
--- a/lib/thrift/ast.ex
+++ b/lib/thrift/ast.ex
@@ -2,8 +2,14 @@ defmodule Thrift.AST do
   @moduledoc """
   Thrift Abstract Syntax Tree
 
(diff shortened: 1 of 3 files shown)


lib/thrift/ast.ex
```diff
@@ -2,8 +2,14 @@ defmodule Thrift.AST do
   @moduledoc """
   Thrift Abstract Syntax Tree
 
-  Parsed Thrift files are repesented as a tree of these structures, starting
-  with a `Thrift.AST.Schema` node.
+  `Thrift.Parser` returns as a tree of these structures, starting with a
```
thecodeboss: "returns a tree"

lib/thrift/ast.ex
```diff
@@ -618,17 +628,4 @@ defmodule Thrift.AST do
...
-          | Include.t()
-          | Constant.t()
-          | TEnum.t()
-          | Field.t()
-          | Exception.t()
-          | Struct.t()
-          | Union.t()
-          | Function.t()
-          | Service.t()
-          | Schema.t()
```
thecodeboss: `parser_test.exs` references this type in one of the specs.

lib/thrift/parser.ex
```diff
@@ -18,7 +21,7 @@ defmodule Thrift.Parser do
   @type opts :: [opt]
 
   @doc """
-  Parses a Thrift document and returns the schema that it represents.
+  Parses a string of Thrift IDL into its AST representation.
```
thecodeboss: Might read better as "Thrift IDL string"

PR #386, opened 2018-08-15
Mark private modules with `@moduledoc false`

jparise:
I don't think any of these modules are intended to be public API, so it
removes their formal @moduledoc documentation. I retained some of that
text as comments because there's some useful information in there.

### DIFF
--- a/lib/thrift.ex
+++ b/lib/thrift.ex
@@ -146,9 +146,10 @@ defmodule Thrift do
     [:bool, :i8, :i16, :i32, :i64, :binary, :double, :byte, :string]
   end
 
-  # A struct for handling IEEE-754 NaN values.
   defmodule NaN do
-    @moduledoc false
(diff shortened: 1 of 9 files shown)


lib/thrift.ex
```diff
@@ -146,11 +146,9 @@ defmodule Thrift do
     [:bool, :i8, :i16, :i32, :i64, :binary, :double, :byte, :string]
   end
 
+  # A struct for handling IEEE-754 NaN values.
   defmodule NaN do
-    @moduledoc """
-    A struct for handling IEEE-754 NaN values.
-    """
-
+    @moduledoc false
```
thecodeboss: This is the only module in this PR I'm a bit iffy on. If someone gets `%Thrift.NaN{}` showing up in their stacktraces, they'll be inclined to go check our docs and won't find this. Could we leave this one in?

PR #387, opened 2018-08-15
Ignore the .elixir_ls directory

jparise:
This is used by the Elixir Language Server and shouldn't be stored in
source control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /_build
+/.elixir_ls
 /cover
 /deps
 /doc


jparise:
> I have this in my global gitignore already.

I have it in mine, too. Do you think we should just assume people will handle this themselves?

thecodeboss:
> I have it in mine, too. Do you think we should just assume people will handle this themselves?

I think we should add it to our gitignore file, I can't see any harm coming from it.

PR #389, opened 2018-08-15
Use a custom `coveralls.safe_travis` task

jparise:
The standard `coveralls.travis` task will unfortunately fail the entire
build if it fails to report to coveralls.io. This isn't the behavior we
want, and until excoveralls grows the ability to do this itself, we'll
use a custom mix task to safely wrap `coveralls.travis`.

I borrowed this idea from benchee.

See: parroty/excoveralls#112


lib/mix/tasks/coveralls.ex
```diff
@@ -0,0 +1,15 @@
...
+  @moduledoc false
+
+  use Mix.Task
+
+  @preferred_cli_env :test
+  @shortdoc "A safe `coveralls.travis` variant that doesn't crash on failed upload."
+
+  def run(args) do
+    Mix.Tasks.Coveralls.do_run(args, type: "travis")
+  rescue
```
thecodeboss: Woah did not know you could [use `rescue` in a `def`](https://hexdocs.pm/elixir/Kernel.html#def/2-rescue-catch-after) like this! Super cool, I think I have some refactors to do.

PR #393, opened 2018-08-19
Document the Constant and TEnum AST types

jparise:
This change also improves the TEnum type specifications:
- The initial values map is required
- Values can only be non-negative integer values

Lastly, mention the common `line` and `annotations` fields in the
top-level AST documentation.

### DIFF
--- a/lib/thrift/ast.ex
+++ b/lib/thrift/ast.ex
@@ -10,6 +10,14 @@ defmodule Thrift.AST do
   - `Thrift.AST.Include`
   - `Thrift.AST.Namespace`
 
+  ## Definitions
+
+  - `Thrift.AST.Constant`
+  - `Thrift.AST.TEnum`
+
+  Each struct has a `line` field that indicates the source line number on
+  which it was defined. Many structs also have an `annotations` field which
+  contains the map of annotations associated with it.
   """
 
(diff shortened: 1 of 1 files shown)


lib/thrift/ast.ex
```diff
@@ -66,25 +76,27 @@ defmodule Thrift.AST do
...
     end
   end
 
   defmodule TEnum do
-    @moduledoc false
-    @type enum_value :: bitstring | integer
+    @moduledoc """
+    An enumerated type with named values.
+    """
+    @type value :: non_neg_integer()
```
thecodeboss: Is this true? Does thrift enforce enum values to be positive? So I couldn't do:
```
enum Weirdo {
  CATS = -1
  DOGS = 0
}
```

jparise: That's what the [Thrift IDL spec](https://thrift.apache.org/docs/idl#enum) says:

> An enum creates an enumerated type, with named values. If no constant value is supplied, the value is either 0 for the first element, or one greater than the preceding value for any subsequent element. Any constant value that is supplied must be non-negative.

... although the Apache Thrift compiler just reads the value as an `int32_t` without a range check. I'm fine with changing this back to `integer()` because we don't have any restrictions on the value range ourself.

PR #394, opened 2018-08-19
Initial service client documentation

jparise:
This moves the service client documentation out of README.md and splits
it between some top-level information in the Thrift module and detailed
documentation in the Thrift.Binary.Framed.Client module. Some of the
text has been revised in the process, but the overall coverage should
remain the same.


scohen:
While i understand why you'd do this, I think that having documentation in the `README.md` file makes it so much easier for people browsing the repo to figure out how to use it (indeed, I've used it a bunch). When landed, this change effectively removes docs since there's no hex package published. 

Also, moving docs into separate module-level documentation spreads them around and makes them harder to navigate, thought the changes to `groups_for_modules` mitigates this a bit.

Instead, you might want to consider adding a `usage` page to the documentation and linking from it on the top level docs.

jparise:
Thanks for reading through this!

> When landed, this change effectively removes docs since there's no hex package published.

This is something we plan on addressing shortly (by publishing the new package and docs).

> moving docs into separate module-level documentation spreads them around and makes them harder to navigate

I've tried to be conscious of this by moving most documentation to the top-level `Thrift` module and moving the details to linked sub-modules, modeled after [Ecto's documentation](https://hexdocs.pm/ecto/Ecto.html).

> you might want to consider adding a usage page to the documentation and linking from it on the top level docs.

Is the advantage to that in keeping a separate markdown file in the repository? Once I bring in the server documentation, I was hoping that this kind of "usage" documentation would all be in `Thrift`, perhaps supplemented by some example projects as suggested in #288.

Alternatively, should we just keep a copy of all of this stuff in the README.md? My view was that we should limit it to just high-level setup and project information.

scohen:
> This is something we plan on addressing shortly (by publishing the new package and docs). 

What's the time frame? We lose docs immediately, and I don't know what the priority is; 2.0 has been on deck for a long while.

> Is the advantage to that in keeping a separate markdown file in the repository? 

Yes, exactly. 

> Alternatively, should we just keep a copy of all of this stuff in the README.md?

That's an option, though you'd have to keep the top-level and module-level docs in sync, which might be a gotcha in the future. 

I also think that in a project such as this, it's helpful to see what the generated server / client looks like, so you can compare it to other alternatives for clarity and ergonomics.

Just my $0.02

jparise:
> What's the time frame?

There isn't a fixed schedule. As documentation is one the last milestone items before releasing it, I've been trying to make progress on this front in my free time.

> I also think that in a project such as this, it's helpful to see what the generated server / client looks like, so you can compare it to other alternatives for clarity and ergonomics.

I agree with this, and it's why we have those examples in the module documentation.

It sounds like we need to decide, overall, if we should put the majority of our effort into module documentation or dedicated text files. I don't have a strong opinion, but as our module documentation was pretty lacking, I've been focusing on that.

jparise:
I've thought about this a little more. It seems like we're discussing three "classes" of documentation. I'll annotate them with where I think they go.

1. Detailed documentation often tied directly to methods (e.g. keyword options). `[@moduledoc]`
2. Guides that explain how to do something: set up the compiler, use a client, etc. `[README.md, Guide.md, ...]`
3. High-level overview documentation that explains the project, what it does, and how it works. `[Thrift module, README.md]`

Do we agree that we're primarily discussing (2) and that we're on the right track with (1) and (3)? If so, I think we're aligned.

thecodeboss:
> 1. Detailed documentation often tied directly to methods (e.g. keyword options). `[@moduledoc]`

I agree with this 100% - we don't need to be documenting keyword arguments at the root readme level.

> 2. Guides that explain how to do something: set up the compiler, use a client, etc. `[README.md, Guide.md, ...]`

Also very important. This PR currently removes all the client docs from the readme, but I think we should keep a minimal client usage example, and then provide a link to the `Client` module for detailed docs on keyword options, ie. (1).

> 3. High-level overview documentation that explains the project, what it does, and how it works. `[Thrift module, README.md]`

We should also have some of this. I like @scohen's idea of having an example of what a generated client/server looks like. Not sure how succinct it would be though

scohen:
@jparise I agree with your analysis; and I think, if we can structure the docs along those lines, I'll be on board. 
Also, if you merge this PR _just_ before 2.0 is released, my concerns will be alleviated.

jparise:
I restored most of the "Using the Client" documentation in the README.md. We can revise the overall README.md file in another pass.

lib/thrift.ex
```diff
@@ -108,10 +108,38 @@ defmodule Thrift do
...
-  same trick for defining Scala namespaces.
+  This alternate syntax is [borrowed from Scrooge][scrooge-namespaces], which
+  uses the same trick for defining Scala namespaces.
 
-  [scrooge-ns]: https://twitter.github.io/scrooge/Namespaces.html
+  [scrooge-namespaces]: https://twitter.github.io/scrooge/Namespaces.html
+
+  ## Clients
+
+  Services clients are built on `Thrift.Binary.Framed.Client`. This module
```
thecodeboss: s/Services/Service/

README.md
```diff
@@ -76,53 +74,6 @@ iex> {:ok, user} = Client.get_user_by_id(client, 22451)
...
-                tcp_opts: [],
-                ssl_opts: [enabled: true, cacertfile: "cacerts.pem", certfile: "cert.pem", keyfile: "key.pem"],
-                gen_server_opts: [timeout: 10_000])
-
-```
-These options set the GenServer timeout to be ten seconds, which means the remote
-side can take its time to reply.
-
-
 ## Using The Server
```
thecodeboss: Since this is not on Hex yet, maybe this doesn't make sense. But we should provide a link to the `Client` docs for users seeking more info on options.

jparise: I agree, but let's wait until we get some docs published. I think we can do that as 2.0.0-rc1 once we finish up the server documentation.

PR #395, opened 2018-08-20
Increase the required Elixir version to 1.5

jparise:
This lets us use `@impl` to formally declare functions that implement
behaviour callbacks (updated here). It also gives us support for
`child_spec/1`-based supervisors (to do).

Elixir 1.5 is the last version that supports OTP 18. Dropping support
for Elixir 1.4 also lets us reduce one row from our testing matrix.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,8 +6,6 @@ branches:
 
 matrix:
   include:
-    - elixir: 1.4
-      otp_release: 19.3
     - elixir: 1.5
       otp_release: 19.3
     - elixir: 1.6
--- a/lib/thrift/binary/framed/client.ex
+++ b/lib/thrift/binary/framed/client.ex
@@ -61,6 +61,7 @@ defmodule Thrift.Binary.Framed.Client do
   require Logger
   use Connection
 
+  @impl Connection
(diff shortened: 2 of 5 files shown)


jparise:
This is also worth a little debate. Moving from 1.4 to 1.5 doesn't give us that many things, but it could be a good thing to do prior to releasing 2.0 given that we don't have any strong backwards compatibility requirements.

I would be most exciting about moving us all the way to Elixir 1.6, but that might be *too* recent for some users.

scohen:
I don't have strong feelings about 1.4 support, what happens if you compile the `@impl` code under 1.4? Something tells me they're quietly dropped. 

Have you updated to 1.6 internally yet?

thecodeboss:
> what happens if you compile the `@impl` code under 1.4?

`@impl` is ignored by pre-1.5 Elixir versions.

I think 1.5 is a good choice. Beyond that is a bit too recent imo

fishcakez:
I think it is okay to require 1.5+ (or 1.6+) on a version `2.0` as it is a breaking change release, and picking a modern version allows us to start with modern approach.

PR #397, opened 2018-08-22
Add module documentation for the server

jparise:
This adds some high-level documentation to the `Thrift` module and
detailed documentation to `Thrift.Binary.Framed.Server`.

### DIFF
--- a/lib/thrift.ex
+++ b/lib/thrift.ex
@@ -140,6 +140,42 @@ defmodule Thrift do
   [naming], so `getUser` becomes `get_user`.
 
   [naming]: http://elixir-lang.org/docs/stable/elixir/naming-conventions.html
+
+  ## Servers
+
+  Thrift servers are a little more involved because you need to create a module
+  to handle the work. Fortunately, a `Behaviour` is generated for each server
+  (complete with typespecs). Just use the `@behavior` module attribute, and the
+  compiler will tell you about any functions you might have missed.
+
+      defmodule UserServiceHandler do
+        @behaviour Thrift.Test.UserService.Handler
+
+        def ping, do: true
+
+        def get_user(user_id) do
+          case Backend.find_user_by_id(user_id) do
+            {:ok, user} ->
+              user
+            {:error, _} ->
+              raise Thrift.Test.UserNotFound.exception message: "could not find user with id #{user_id}"
+          end
+        end
+
+        def delete(user_id) do
+          Backend.delete_user(user_id) == :ok
+        end
+      end
+
+  To start the server:
+
+      {:ok, pid} = Thrift.Test.UserService.Binary.Framed.Server.start_link(UserServiceHandler, 2345, [])
+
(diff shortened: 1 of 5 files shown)


lib/thrift/binary/framed/server.ex
```diff
@@ -8,10 +14,45 @@ defmodule Thrift.Binary.Framed.Server do
...
+
+  `ssl_opts`: A keyword list of SSL/TLS options:
+
+  - `:enabled`: A boolean indicating whether to upgrade the connection to
+    the SSL protocol
+  - `:optional`: A boolean indicating whether to accept both SSL and plain
+    connections
+  - `:configure`: A 0-arity function to provide additional SSL options at
+    runtime
+  - Additional `t:Thrift.Transport.SSL.option/0` values specifying other
```
fishcakez: The additional options are `:ssl.ssl_option`?

lib/thrift.ex
```diff
@@ -140,6 +140,42 @@ defmodule Thrift do
   [naming], so `getUser` becomes `get_user`.
 
   [naming]: http://elixir-lang.org/docs/stable/elixir/naming-conventions.html
+
+  ## Servers
+
+  Thrift servers are a little more involved because you need to create a module
+  to handle the work. Fortunately, a `Behaviour` is generated for each server
+  (complete with typespecs). Just use the `@behavior` module attribute, and the
```
fishcakez: s/Just use/Use/

lib/thrift/binary/framed/server.ex
```diff
@@ -8,10 +14,45 @@ defmodule Thrift.Binary.Framed.Server do
...
 
   @type server_opts :: [server_option]
 
-  @spec start_link(module, 1..65_535, module, [server_option]) :: GenServer.on_start()
+  @doc """
+  Starts the server using the specified handler module.
+
+  The following server options can be specified:
+
+  `:worker_count`: The number of acceptor workers used to take requests.
```
fishcakez: acceptor workers to accept on the socket, different processes are used to handle requests

lib/thrift/binary/framed/server.ex
```diff
@@ -8,10 +14,45 @@ defmodule Thrift.Binary.Framed.Server do
...
+
+  `:name`: (Optional) The name of the server. The server's pid becomes
+  registered under this name. If not specified, the handler module's name
+  is used.
+
+  `:max_restarts`: The number of times to restart (see `:max_seconds`).
```
fishcakez: I don't think we need to expose these. Instead set `max_restarts: 0` and allow the parent supervisor to handle things. The supervisor this configures is hidden from the user.

jparise: That sounds like more than a documentation, yes? I'll remove these from the documentation here, and we can do the other bits separately.

jparise: Done in #400.

PR #403, opened 2018-08-23
Allow annotation after field type

fishcakez:
Other thrift libraries allow annotations in a few more places. Add support for these in schemas but do not add to the AST.

### DIFF
--- a/src/thrift_parser.yrl
+++ b/src/thrift_parser.yrl
@@ -159,8 +159,8 @@ Throws -> throws '(' FieldList ')': '$3'.
 FieldList -> '$empty': [].
 FieldList -> Field FieldList: ['$1'|'$2'].
 
-Field -> FieldIdentifier FieldRequired FieldType ident FieldDefault Annotations Separator:
-    build_node('Field', line('$4'), '$6', ['$1', '$2', '$3', unwrap('$4'), '$5']).
+Field -> FieldIdentifier FieldRequired FieldType Annotations ident FieldDefault Annotations Separator:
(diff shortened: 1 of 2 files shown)


src/thrift_parser.yrl
```diff
@@ -159,8 +159,8 @@ Throws -> throws '(' FieldList ')': '$3'.
 FieldList -> '$empty': [].
 FieldList -> Field FieldList: ['$1'|'$2'].
 
-Field -> FieldIdentifier FieldRequired FieldType ident FieldDefault Annotations Separator:
-    build_node('Field', line('$4'), '$6', ['$1', '$2', '$3', unwrap('$4'), '$5']).
+Field -> FieldIdentifier FieldRequired FieldType Annotations ident FieldDefault Annotations Separator:
```
fishcakez: Annotation following the `FieldType` is supported by apache and fbthrift. This can be used to signal the type of map/set/list.

src/thrift_parser.yrl
```diff
@@ -190,8 +190,8 @@ BaseType -> double: double.
 BaseType -> string: string.
 BaseType -> binary: binary.
 
-MapType -> map '<' FieldType ',' FieldType '>': {'$3', '$5'}.
-SetType -> set '<' FieldType '>': '$3'.
+MapType -> map '<' FieldType Annotations ',' FieldType Annotations '>': {'$3', '$6'}.
+SetType -> set '<' FieldType Annotations '>': '$3'.
```
fishcakez: Added these to match `ListType`/`Field`, it seems that annotations can follow field type every where else.

fishcakez:
We may want to increase support for these annotations in futures to have custom elixir/erlang types for these.

PR #405, opened 2018-08-23
Support cpp_include

fishcakez:
Thrift schemas may include cpp_include to include C++ files, add support in the parser.

### DIFF
--- a/test/thrift/parser/parser_test.exs
+++ b/test/thrift/parser/parser_test.exs
@@ -8,6 +8,7 @@ defmodule Thrift.Parser.ParserTest do
 
   alias Thrift.AST.{
     Constant,
+    CppInclude,
     Exception,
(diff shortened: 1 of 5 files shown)


test/thrift/parser/parser_test.exs
```diff
@@ -111,6 +112,22 @@ defmodule Thrift.Parser.ParserTest do
            ]
   end
 
+  test "parsing cpp_include headers" do
+    cpp_includes =
+      parse(
+        """
+        cpp_include "foo.h"
+        cpp_include "bar.h"
```
jparise: Perhaps also have a test example that uses slashes, like `”bar/baz.h”`.

fishcakez: Added

PR #406, opened 2018-08-23
Support empty unix comment

fishcakez:
When adding support for custom namespace via unix comments we introduced an edge case for the empty unix comment where the parse include the next line in the comment. Add processing to handle this case.


thecodeboss:
Good find

PR #407, opened 2018-08-24
Fix missing struct key in AST.TypeRef

thecodeboss:
I was inspecting some of these `TypeRef` structs and noticed they were appearing like this:
```elixir
%{
  __struct__: Thrift.AST.TypeRef,
  line: 4,
  name: :"includes.AnotherInteger",
  referenced_type: :"includes.shared.MyInteger"
}
```
instead of
```elixir
%Thrift.AST.TypeRef{
  line: 4,
  name: :"includes.AnotherInteger",
  referenced_type: :"includes.shared.MyInteger"
}
```
This fixes it - they were missing the `:name` key in their struct definition

### DIFF
--- a/lib/thrift/ast.ex
+++ b/lib/thrift/ast.ex
@@ -316,7 +316,7 @@ defmodule Thrift.AST do
     @type t :: %TypeRef{line: Thrift.Parser.line(), referenced_type: atom}
 
     @enforce_keys [:referenced_type]
-    defstruct line: nil, referenced_type: nil
+    defstruct line: nil, name: nil, referenced_type: nil
 
     @spec new(charlist) :: t
     def new(referenced_type) do


lib/thrift/ast.ex
```diff
@@ -316,7 +316,7 @@ defmodule Thrift.AST do
     @type t :: %TypeRef{line: Thrift.Parser.line(), referenced_type: atom}
```
jparise: This should also be updated.

thecodeboss:
> Where is `name` being assigned? I did a quick look through the code, and I'm not seeing anything.

`Thrift.Parser.Resolver` line 30. The resolver adds a `:name` field to typedefs once it resolves the type they refer to.

PR #409, opened 2018-08-24
Fix typedefs defined as references to another typedef

### LINKED ISSUE #408: A typedef which references an included typedef fails to compile
Compiling these for example fails:
#### shared.thrift
```thrift
typedef i32 MyInteger
```
#### test.thrift
```thrift
include "shared.thrift"

typedef shared.MyInteger MyType

struct StructWithInclude {
  1: optional MyType num = 5;
}
```

For reference, the resulting stacktrace looks like:
```elixir
** (RuntimeError) Fatal error: Could not find type: MyType
    (thrift) lib/thrift/generator/utils.ex:345: Thrift.Generator.Utils.resolution_failed/1
    (thrift) lib/thrift/generator/utils.ex:222: Thrift.Generator.Utils.quote_value/3
    (thrift) lib/thrift/generator/struct_generator.ex:20: anonymous fn/2 in Thrift.Generator.StructGenerator.generate/4
    (elixir) lib/enum.ex:1294: Enum."-map/2-lists^map/1-0-"/2
    (elixir) lib/enum.ex:1294: Enum."-map/2-lists^map/1-0-"/2
    (thrift) lib/thrift/generator/struct_generator.ex:18: Thrift.Generator.StructGenerator.generate/4
    (thrift) lib/thrift/generator.ex:188: anonymous fn/3 in Thrift.Generator.generate_struct_modules/1
    (stdlib) lists.erl:1263: :lists.foldl/3
    (thrift) lib/thrift/generator.ex:186: Thrift.Generator.generate_struct_modules/1
    (thrift) lib/thrift/generator.ex:74: Thrift.Generator.generate_schema/1
    (thrift) lib/thrift/generator.ex:51: anonymous fn/3 in Thrift.Generator.generate!/2
    (elixir) lib/enum.ex:1012: anonymous fn/3 in Enum.flat_map/2
    (stdlib) lists.erl:1263: :lists.foldl/3
    (elixir) lib/enum.ex:1915: Enum.flat_map/2
    (elixir) lib/enum.ex:2924: Enum.flat_map_list/2
    (elixir) lib/enum.ex:2925: Enum.flat_map_list/2
    (thrift) test/support/lib/thrift_test_case.ex:93: ThriftTestCase.generate_files/3
    (thrift) test/support/lib/thrift_test_case.ex:42: ThriftTestCase.quoted_contents/2
    test/thrift/generator/models_test.exs:196: (module)
```
I'm currently working on fixing it, but it's proving to be quite tricky. The type resolution code is a bit hairy to follow for nested definitions like this.

thecodeboss:
Fixes #408.

There were two elements of the AST that were having their namespaces handled the same, though they are actually supposed to be different. `TypeRef` and `ValueRef`.

`ValueRef`s refer to enums, and the logic is supposed to be that if its fully-qualified name already has the current module in it (ex: `additions.ChocolateAdditions.ALMONDS`), then we don't re-add the module (as this would result in `additions.additions.ChocolateAdditions.ALMONDS`).

This logic was also being applied to `TypeRef`s, however there's an edge case that it doesn't work on. If a typeref already refers to a qualified type (ex: `typedef shared.Integer MyInteger`), an extra module could get prepended (ex: `current_module.shared.Integer`). This is invalid.

The fix is to break up `add_namespace_to_type` into two functions: `add_namespace_to_type` and `add_namespace_to_value`, and handle this edge case.

I think this makes sense. I fear this introduces more edge cases somehow. I tried to test a bunch of cases I could think of, but hopefully I'm not missing any?

### DIFF
--- a/lib/thrift/ast.ex
+++ b/lib/thrift/ast.ex
@@ -577,8 +577,42 @@ defmodule Thrift.AST do
         |> String.split(".")
 
       case split_type_name do
+        [_module, _type | _rest] ->
+          # this case accounts for types that already have a module in them
+          type_name
+
+        _ ->
+          :"#{module}.#{type_name}"
+      end
+    end
+
+    defp add_namespace_to_value(module, {:set, elem_type}) do
+      {:set, add_namespace_to_value(module, elem_type)}
+    end
+
+    defp add_namespace_to_value(module, {:list, elem_type}) do
+      {:list, add_namespace_to_value(module, elem_type)}
+    end
+
+    defp add_namespace_to_value(module, {:map, {key_type, val_type}}) do
+      {:map, {add_namespace_to_value(module, key_type), add_namespace_to_value(module, val_type)}}
+    end
+
+    for type <- Thrift.primitive_names() do
+      defp add_namespace_to_value(_, unquote(type)) do
+        unquote(type)
+      end
+    end
+
+    defp add_namespace_to_value(module, type_name) when is_atom(type_name) do
+      split_type_name =
+        type_name
+        |> Atom.to_string()
+        |> String.split(".")
+
+      case split_type_name do
         [^module | _rest] ->
-          # this case accounts for types that already have the current module in them
+          # this case accounts for values that already have the current module in them
           type_name
 
         _ ->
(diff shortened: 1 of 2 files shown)


fishcakez:
I think that we should change the data structure for names to a list of atoms. This prevents some edge cases and going atom <> string. Also in our case each part of the name is a separate term so it fits better into the thinking. I think this would also make it so we could always use fully qualified names.

thecodeboss:
That sounds good, perhaps we should wait until 2.0 comes out before refactoring it though.

PR #412, opened 2018-08-28
Add an example project

thecodeboss:
@jparise and I discussed adding this, and it is also mentioned as an item in #288.

The example project is a simple calculator service, of which you'll find more detail below since I added a readme. The readme gets included in generated docs, the example project tests get run as part of the Travis suite, and the formatter also ensures the example is kept formatted.


example/README.md
```diff
@@ -0,0 +1,19 @@
...
+The [`example/`](ex) directory of elixir-thrift contains an implementation of a simple calculator service. It serves two purposes:
+
+1. How to use this library.
+2. Example of what elixir-thrift generated code looks like.
+
+[ex]: https://github.com/pinterest/elixir-thrift/example
+
+## Calculator Service
+
+The service is very simple, and implements the four basic arithmetic operations: addition, subtraction, multiplication, and division. It demonstrates [how to add a thrift server to a supervision tree](supervisor), and [how to write a handler](handler) for a [service defined in thrift](thrift-defs). It also demonstrates how to use exceptions by implementing a division-by-zero exception.
```
jparise: Capitalize Thrift here and below.

example/lib/calculator/service_handler.ex
```diff
@@ -0,0 +1,30 @@
...
+
+  @impl true
+  def multiply(left, right) do
+    left * right
+  end
+
+  @impl true
+  def divide(left, right) do
+    case right do
+      0 ->
```
jparise: Or match on `right==0` (e.g. `divide(left, 0)`)?

thecodeboss:
> Do you think we should also include an enum and struct in the example, too? I wouldn't want to make it too complicated, but those are the other prominent types that people will regularly use.

For structs, perhaps we could define:
```thrift
struct Vector {
  1: double x = 0.0,
  2: double y = 0.0,
  3: double z = 0.0,
}

enum VectorProductType {
  DOT_PRODUCT = 1,
  CROSS_PRODUCT = 2
}
```
and in the service definition:
```thrift
vectorProduct(1: Vector left, 2: Vector right, 3: VectorProductType type)
```
This would cover both structs and enums with examples, while also being a somewhat realistic calculator operation. Thoughts?

thecodeboss:
Actually the vector product has different return types depending on the operation (vector or scalar). Maybe a good chance to demo unions as well

jparise:
> This would cover both structs and enums with examples, while also being a somewhat realistic calculator operation. Thoughts?

That looks good to me. I also thought of using fixed-point notation.

example/lib/calculator/service_handler.ex
```diff
@@ -0,0 +1,29 @@
+defmodule Calculator.ServiceHandler do
```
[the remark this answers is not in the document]
thecodeboss: Go away ebert, I added a moduledoc.

example/lib/calculator/service_handler.ex
```diff
@@ -18,13 +22,34 @@ defmodule Calculator.ServiceHandler do
...
+
   def divide(left, right) do
-    case right do
-      0 ->
-        raise Calculator.Generated.DivideByZeroError, message: "Cannot divide by zero"
+    div(left, right)
+  end
+
+  @impl true
+  def vector_product(left, right, type) do
```
jparise: Could also match on `VectorProductType`. 😉

thecodeboss: This is more tricky since it's a macro. First approach would be
```elixir
def vector_product(left, right, VectorProductType.dot_product())
```
but the compiler will not allow you to do this.

Second approach is:
```elixir
@dot_product VectorProductType.dot_product()

def vector_product(left, right, @dot_product)
```
But this is just plain ugly.
I settled with the final approach here.

thecodeboss: This is a very good example of why #312 would be useful btw

jparise: Oh, that's right. I was thinking we had already done something with #313.

PR #416, opened 2018-08-31
Fix the example/ directory link

jparise:


thecodeboss:
Good catch. I had tried to predict what the links would be, but got it wrong.

There are 4 other links at the bottom of the readme, can you update those too?

PR #436, opened 2018-10-17
Store FileGroup namespace values as strings

jparise:
We previously stored a mapping of `module_name -> %Namespace{}`, but we
only need the namespace's value to resolve the destination module.

Just storing the namespace values directly (in their native string form)
simplifies this code, including the "default namespace" logic (because
it no longer needs to be represented as a full `%Namespace{}`).

### DIFF
--- a/test/thrift/parser/parser_test.exs
+++ b/test/thrift/parser/parser_test.exs
@@ -918,17 +918,14 @@ defmodule Thrift.Parser.ParserTest do
 
   describe "namespace option" do
     setup do
-      path = Path.join(@test_file_dir, "namespace.thrift")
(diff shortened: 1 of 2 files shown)


test/thrift/parser/parser_test.exs
```diff
@@ -918,17 +918,14 @@ defmodule Thrift.Parser.ParserTest do
 
   describe "namespace option" do
     setup do
-      path = Path.join(@test_file_dir, "namespace.thrift")
+      path = Path.join(@test_file_dir, "module.thrift")
```
jparise: I renamed this to make the test code below a little clearer. There were too many things named "namespace" before.

PR #440, opened 2018-11-22
Add exception return type to client typespec

thecodeboss:
When doing `Client.call(...)`, it's possible it will return a `{:error, {:exception, struct}}` tuple (and many of our tests confirm this).

However the type definition does not reflect this, and so Dialyzer complains on generated code saying the `{:error, {:exception, ex}}` clause will never match `{:error, atom}`.


thecodeboss:
I personally prefer this style instead of the union because it makes the spec appear less confusing. But it's also not a big deal; no one seems opposed so I'll merge.

PR #443, opened 2018-12-03
Remove "include name" concept from FileRef

jparise:
First, remove `FileRef.include_name/1` and the struct field of the same
name. These were really never being used. Instead, we can construct the
`ParsedFile.name` field value directly.

Also, remove the `FileRef.thrift_include` type and just use `Path.t()` in its
place. This is also more correct than the former's `String.t()` spec.


jparise:
I believe we can actually eliminate `FileRef` entirely, but that will involve a larger follow-up change.

Issue #236, opened 2017-02-11
RFC:  Re-evaluate the parser/generator wrt multiple write-outs.


dantswain:
This is a follow-up to #227 (fixed by #230).  The current behavior of the parser/generator is to iterate over the list of input files and do a full parse/generate for each.  This leads to a fair amount of double-processing and at times writing out generated code more than once.  For the most part this isn't a huge problem because the generated code is consistent, but as we saw with Constants it _can_ lead to modules being clobbered.

On one hand, it seems like the right thing to do might be to iterate over the list of input files (and any files that those include) and parse out one unified schema from which to generate code.  This should be the most consistent way to generate output.  It seems like it would also require fairly significant changes to the code.

On the other hand, I believe the way we are doing it now is close(r) to the way that most of the official thrift generators work: generally one executes `thrift -o whateverlanguage path/to/some.thrift` which parsers a single input file and generates code from that file.  I'm not 100% sure whether that will generate structs/enums/etc from files that are `include`d from the original file.  These parsers don't seem to really go out of their way to make sure that developers don't clobber their own generated code due to name collisions.  I'm also not sure that they go out of their way to make sure that any code you generate will be complete from a name resolution standpoint (i.e., if you reference `Foo`, it may not always care if you've processed the file that generates `Foo`).

An intermediate solution would be to only generate code for the "initial file" in the file group.  I.e., if file `x.thrift` does `include "y.thrift"` which defines `struct Y`, then we parse `y.thrift` to resolve references but do not generate `Y` except when specifically processing `y.thrift`.

I guess a first step here would be to do some research on how the existing parsers work, especially wrt how they handle `include`d files - I'm kind of shooting from the cuff based on past experience but I'm not 100% confident.

jparise:
I think this is the right direction for the parser and generation. I'll own this for the time being, but I don't expect to work on it until at least the [2.1 milestone](https://github.com/pinterest/elixir-thrift/milestone/2).

PR #447, opened 2018-12-09
Rework the Thrift.Parser public interface

jparise:
The primary goal of this work is to provide a better interface for
surfacing parser errors to the compiler task, but this change also
simplifies the calling conventions and makes the code a bit more
idiomatic.

Specifically:

 - The `ParsedFile` module is gone. The `Thrift.Parser` functions now do
   all of the parsing work.
 - `parse/1` has been renamed `parse_string/1`.
 - `parse_file/1` now mirrors `parse_string/1` and returns the parsed
   schema or an error tuple.
 - `parse_file_group/2` and `parse_file_group!/2` have been added for
   recursively parsing a Thrift file and its included files.


lib/thrift/parser.ex
```diff
@@ -41,26 +43,76 @@ defmodule Thrift.Parser do
...
 
-  In addition to the given file, all included files are also parsed and
-  returned as part of the resulting `Thrift.Parser.FileGroup`.
+  @doc """
+  Parses a Thrift IDL file and its included files into a file group.
   """
-  @spec parse_file(Path.t(), opts) :: FileGroup.t()
-  def parse_file(file_path, opts \\ []) do
+  @spec parse_file_group(Path.t(), opts) :: {:ok, FileGroup.t()} | error
+  def parse_file_group(path, opts \\ []) do
```
fishcakez: Do we want to keep the module dependency parser -> file group now that we have a `parse_file` that returns AST? We require file group -> parser dependency to handle includes so have an existing circular dependency. Its quite awkward as `FileGroup` takes AST as input but then does its own parsing on files for includes. For a cleaner API I think we should always use only one type of input. As we need to parse included files in `FileGroup` the input would be paths. Therefore I think that we should resolve this by moving this function to `FileGroup` and only supporting paths as input to `FileGroup` functions, and `FileGroup` would then call `Parser`.

This should also make it cleaner to support multiple root schemas for a single file group with common files that we discussed.

jparise: Yes, I started thinking about this a little in context #236. I don't have any strong feelings yet, but I agree it's going to take a bit of thought.

PR #448, opened 2018-12-13
(WIP) Initial work on nimble_parsec generated lexer

fishcakez:
Initial work on a nimble parsec based lexer. Tried to design for nice errors and to avoid back tracking on happy path.

### DIFF
--- a/lib/thrift/parser/nimble.ex
+++ b/lib/thrift/parser/nimble.ex
@@ -1,0 +1,189 @@
+defmodule Thrift.Parser.Nimble do
+  import NimbleParsec
+
+  defmodule Lexer do
+    @moduledoc false
+    @punctuator ~c"(){}[]<>,;=*"
+
+    @keywords ~w(
+      namespace include cpp_include
+      typedef enum union struct exception
+      void bool byte i8 i16 i32 i64 double string binary list map set
+      const oneway extends throws service required optional
+      true false
+    )
+
+    def token(combinator \\ empty()) do
+      choice(combinator, [
+        keyword(),
+        identifier(),
+        literal(),
+        number(),
+        punctuator(),
+        whitespace()
+      ])
+    end
+
+    defp whitespace() do
+      ascii_string([?\s, ?\t, ?\n, ?\v, ?\f, ?\r], min: 1)
+      |> ignore()
+      |> label("whitespace")
+    end
+
+    defp keyword() do
+      @keywords
+      |> Enum.map(&(string(&1) |> replace(String.to_atom(&1))))
+      |> choice()
+      |> label("keyword")
+    end
+
+    defp punctuator() do
+      @punctuator
+      |> Enum.map(&(string(<<&1>>) |> replace(String.to_atom(<<&1>>))))
+      |> choice()
+      |> label("punctuator")
+    end
+
+    defp literal() do
+      choice([
+        literal_with(?"),
+        literal_with(?')
+      ])
+      |> reduce({List, :to_string, []})
+      |> label("literal")
+    end
+
+    defp literal_with(char) do
+      delim = ascii_char([char])
+
+      delim
+      |> ignore()
+      |> concat(
+        choice([
+          utf8_char([?\\]) |> ignore() |> concat(delim),
+          utf8_char([]),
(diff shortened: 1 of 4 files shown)


lib/thrift/parser/nimble.ex
```diff
@@ -0,0 +1,189 @@
...
+    defp literal() do
+      choice([
+        literal_with(?"),
+        literal_with(?')
+      ])
+      |> reduce({List, :to_string, []})
+      |> label("literal")
+    end
+
+    defp literal_with(char) do
```
jparise: Perhaps `quoted_literal(quote_char)`?

lib/thrift/parser/nimble.ex
```diff
@@ -0,0 +1,189 @@
...
+    end
+
+    defp number(combinator \\ empty()) do
+      combinator
+      |> choice([
+        ascii_char([?-, ?+])
+        |> choice([
+          unsigned_number(),
+          empty()
+          |> error("expected number")
```
jparise: I think this reads a little better as (like you do above):

```elixir
error(empty(), "expected number")
```

lib/thrift/parser/nimble.ex
```diff
@@ -0,0 +1,189 @@
...
+
+        [exponent, exponent_sign, fraction, int] ->
+          {[String.to_float("#{int}.#{fraction}e#{[exponent_sign]}#{exponent}")], context}
+      end
+    end
+
+    def identifier() do
+      ascii_char([?a..?z, ?A..?Z, ?_])
+      |> repeat(ascii_char([?a..?z, ?A..?Z, ?_, ?0..?9]))
+      |> reduce({List, :to_atom, []})
```
jparise: Consider extracting these three lines into their own function because we repeat them below.

lib/thrift/parser/nimble.ex
```diff
@@ -0,0 +1,189 @@
+defmodule Thrift.Parser.Nimble do
```
jparise: Do you think we'll have multiple parsecs in here? Otherwise, just `Thrift.Parser.Lexer` (`lib/thrift/parser/lexer.ex`) seems better.

mix.exs
```diff
@@ -83,6 +83,10 @@ defmodule Thrift.Mixfile do
       {:credo, "~> 1.0", only: :dev, runtime: false},
       {:dialyxir, "~> 0.5", only: :dev, runtime: false},
 
+      # Compile
+      {:nimble_parsec, "~> 0.4",
```
jparise: [0.5](https://github.com/plataformatec/nimble_parsec/releases/tag/v0.5.0) was just released.

lib/thrift/parser/nimble.ex
```diff
@@ -0,0 +1,189 @@
...
+    end
+
+    defp literal_with(char) do
+      delim = ascii_char([char])
+
+      delim
+      |> ignore()
+      |> concat(
+        choice([
+          utf8_char([?\\]) |> ignore() |> concat(delim),
```
jparise: Does this handle embedded newlines, etc. like we support in the current lexer?

https://github.com/pinterest/elixir-thrift/blob/1610f612e8638307e9b98970d35839299facf5f1/src/thrift_lexer.xrl#L91-L105

scohen:
@jparise I'm curious why those reserved words are reserved. What was the idea behind them?

jparise:
> @jparise I'm curious why those reserved words are reserved. What was the idea behind them?

This is borrowed from Apache Thrift. The idea is that Thrift IDL shouldn't contain words that would clash with an output language's keywords when code generated. There are two possible approaches:

1. Prevent the IDL from _ever_ containing languages' keywords (across a large set of supported languages).
2. Let the IDL contain close to anything, and deal with reserved words on a per-generator basis. The generator would have to recognize its own reserved words and renamed them accordingly.

Apache Thrift generally does (1) (at least for C++'s reserved words), so we copied that behavior in our lexer for compatibility. In practice, we should only be concerned with Elixir's reserved words (which aren't much of a problem) and could take approach (2).

In this project, reserved word rejection was added in #295 by @thecodeboss.

scohen:
Makes sense, it was a bit perplexing to see keywords for other languages in here.

PR #460, opened 2019-03-17
Return diagnostics from our compile.thrift task

jparise:
This improves our support for the [`Mix.Task.Compiler` behaviour](https://hexdocs.pm/mix/Mix.Task.Compiler.html) by returning a list of diagnostics representing any parser errors that occurred while processing the input files.

Generator errors are not yet converted to diagnostics. That will come next.

Errors are still printed to the shell, but they are now derived from the [`Mix.Task.Compiler.Diagnostic`](https://hexdocs.pm/mix/Mix.Task.Compiler.Diagnostic.html) structure. This provides a nice abstraction and presented the opportunity to change the message format to better match the output of the Elixir compiler.

Lastly, the unit tests have been rewritten to use a new `MixTest.Case` template. Mix shell output is now routed through [`Mix.Shell.Process`](https://hexdocs.pm/mix/Mix.Shell.Process.html) so we no longer need to use `ExUnit.CaptureIO`, and the common fixture-related helper routines are now shared.


lib/mix/tasks/compile.thrift.ex
```diff
@@ -68,21 +69,29 @@ defmodule Mix.Tasks.Compile.Thrift do
...
+        {groups, [] = _diagnostics} ->
+          groups
+          |> extract_targets(output_path, opts[:force])
+          |> generate(manifest(), output_path, opts)
+
+        {_groups, diagnostics} ->
+          {:error, diagnostics}
+      end
+
+    Enum.each(diagnostics, &print_diagnostic/1)
```
fishcakez: Would it be nicer user experience if we could print errors as they happen? I think this makes the abstraction in the code less nice. Perhaps we could revisit that later.

lib/mix/tasks/compile.thrift.ex
```diff
@@ -91,13 +100,19 @@ defmodule Mix.Tasks.Compile.Thrift do
...
-  defp parse(file, opts) do
-    Thrift.Parser.parse_file_group!(file, opts)
-  rescue
-    e ->
-      Mix.shell().error("Failed to parse #{file}: #{Exception.message(e)}")
-      nil
+  @spec parse([Path.t()], Thrift.Parser.opts()) :: {[FileGroup.t()], [Diagnostic.t()]}
+  defp parse(files, opts) do
+    files
+    |> Enum.reverse()
```
fishcakez: The reverse here is so the output is same order as input? Or is there something deeper?

jparise: That's correct. It's so we can push the following diagnostics onto the head of the `diagnostics` list and the result remains consistent with the ordering of the input files.

lib/mix/tasks/compile.thrift.ex
jparise: I figured there were three options:

1. **Print the errors as they happen in the Parser / Generator code.** This has the dual disadvantage of decentralizing error output and emitting output from functions outside of the mix context.
2. **Print errors as we receive them in the mix task.** I think this is what you're suggesting, and I think it would work fine but at the cost of spreading out the code a bit. We could still use `Diagnostic` as the normalized input structure into the error-printing code by converting all errors/warnings to `Diagnostic` first.
3. **Print errors as the end of the stream.** This felt like the simplest implementation because we're guaranteed that all diagnostics have been collected at this point. That's the approach you're seeing here.

I think (2) has some advantages because the error output is reported sooner, but I also felt like most error cases abort the compilation process early anyway, so in practice there might not be much of a delay between an error occurring and being reported.

thecodeboss: I think most cases aborting compilation is a solid argument for going with option (3).

PR #478, opened 2019-09-16
allow client to reconnect if other side closes connection

cybernetlab:
It would be nice to allow framed client to reconnect if other side closed connection because of client inactivity for example. In my personal case thrift library used as a client for `hbase` and this server closes inactive connections in about 10 minutes. Also I'm using `poolboy` to pool connections and without reconnection ability I'll need to do some unnecessary logic to monitor status of tcp connections and push out clients with closed connection from the pool.

### DIFF
--- a/lib/thrift/binary/framed/client.ex
+++ b/lib/thrift/binary/framed/client.ex
@@ -42,6 +42,7 @@ defmodule Thrift.Binary.Framed.Client do
           {:tcp_opts, [tcp_option]}
           | {:ssl_opts, [SSL.option()]}
           | {:gen_server_opts, [genserver_call_option]}
+          | {:reconnect, boolean}
 
   @type options :: [option]
 
@@ -55,7 +56,8 @@ defmodule Thrift.Binary.Framed.Client do
             ssl_opts: [SSL.option()],
             timeout: integer,
             sock: {:gen_tcp, :gen_tcp.socket()} | {:ssl, :ssl.sslsocket()},
-            seq_id: integer
+            seq_id: integer,
+            reconnect: boolean
           }
     defstruct host: nil,
               port: nil,
@@ -64,7 +66,8 @@ defmodule Thrift.Binary.Framed.Client do
               ssl_opts: nil,
               timeout: 5000,
               sock: nil,
-              seq_id: 0
+              seq_id: 0,
+              reconnect: false
   end
 
   require Logger
@@ -74,6 +77,7 @@ defmodule Thrift.Binary.Framed.Client do
   def init({host, port, opts}) do
     tcp_opts = Keyword.get(opts, :tcp_opts, [])
     ssl_opts = Keyword.get(opts, :ssl_opts, [])
+    reconnect = Keyword.get(opts, :reconnect, false)
 
     {timeout, tcp_opts} = Keyword.pop(tcp_opts, :timeout, 5000)
 
@@ -82,7 +86,8 @@ defmodule Thrift.Binary.Framed.Client do
       port: port,
       tcp_opts: tcp_opts,
       ssl_opts: ssl_opts,
-      timeout: timeout
+      timeout: timeout,
+      reconnect: reconnect
     }
 
     {:connect, :init, s}
@@ -246,7 +251,7 @@ defmodule Thrift.Binary.Framed.Client do
   def handle_call(
         {:call, rpc_name, serialized_args, tcp_opts},
         _,
-        %{sock: {transport, sock}, seq_id: seq_id, timeout: default_timeout} = s
+        %{sock: {transport, sock}, seq_id: seq_id, timeout: default_timeout, reconnect: reconnect} = s
       ) do
     s = %{s | seq_id: seq_id + 1}
     message = Binary.serialize(:message_begin, {:call, seq_id, rpc_name})
@@ -257,6 +262,13 @@ defmodule Thrift.Binary.Framed.Client do
       reply = deserialize_message_reply(message, rpc_name, seq_id)
       {:reply, reply, s}
     else
+      {:error, :closed} = error ->
+        if reconnect do
+          {:connect, :reconnect, s}
+        else
+          {:disconnect, error, error, s}
+        end
+
       {:error, :timeout} = error ->
         {:disconnect, {:error, :timeout, timeout}, error, s}
 
@@ -277,7 +289,7 @@ defmodule Thrift.Binary.Framed.Client do
 
   def handle_cast(
         {:oneway, rpc_name, serialized_args},
-        %{sock: {transport, sock}, seq_id: seq_id} = s
+        %{sock: {transport, sock}, seq_id: seq_id, reconnect: reconnect} = s
       ) do
     s = %{s | seq_id: seq_id + 1}
     message = Binary.serialize(:message_begin, {:oneway, seq_id, rpc_name})
@@ -285,6 +297,13 @@ defmodule Thrift.Binary.Framed.Client do
     case transport.send(sock, [message | serialized_args]) do
       :ok ->
         {:noreply, s}
+
+      {:error, :closed} = error ->
+        if reconnect do
+          {:connect, :reconnect, s}
+        else
+          {:disconnect, error, s}
+        end
 
       {:error, _} = error ->
         {:disconnect, error, s}


cybernetlab:
Ah, sorry - I've forgotten to resend data after connection - some addition fixes needed

cybernetlab:
Now all works fine in my environment. What should I do with test coverage? )

lib/thrift/binary/framed/client.ex
```diff
@@ -372,4 +401,34 @@ defmodule Thrift.Binary.Framed.Client do
...
+    case handle_call(msg, from, s) do
+      {:reply, reply, s} ->
+        GenServer.reply(from, reply)
+        {:ok, s}
+
+      {:disconnect, info, error, s} ->
+        GenServer.reply(from, error)
+        disconnect(info, s)
+
+      _ ->
```
fishcakez: The handle call could return `{:connect, _, _, _}` here, and we seem to do the wrong thing by returning an ok socket.

For example it could be possible due to a server side bug that the server always closes the connection without sending a response for certain requests - and we would always be seeing it.

lib/thrift/binary/framed/client.ex
```diff
@@ -372,4 +401,34 @@ defmodule Thrift.Binary.Framed.Client do
...
+
+  defp maybe_resend_data({:ok, s}, {:reconnect, :cast, msg}) do
+    case handle_cast(msg, s) do
+      {:noreply, s} ->
+        {:ok, s}
+
+      {:disconnect, info, s} ->
+        disconnect(info, s)
+
+      _ ->
```
fishcakez: Same here?

lib/thrift/binary/framed/client.ex
```diff
@@ -286,6 +308,13 @@ defmodule Thrift.Binary.Framed.Client do
       :ok ->
         {:noreply, s}
 
+      {:error, :closed} = error ->
+        if reconnect do
+          {:connect, {:reconnect, :cast, msg}, %{s | sock: nil}}
```
fishcakez: We should always go through the disconnect -> connect cycle to ensure we cleanup and dont miss anything in disconnect/2. Its a little awkward but means we need to pass the reconnect info through disconnect -> connect.

lib/thrift/binary/framed/client.ex
cybernetlab: Actually this is a return from `connect` callback - not from `handle_call`. The result of call to `disconnect(info, s)` returned from `connect` callback [here](https://github.com/cybernetlab/elixir-thrift/blob/allow-reconnect/lib/thrift/binary/framed/client.ex#L158) and it can be only `{:ok, _}` or `{:stop, _}` (or `{:backoff, _}` which is not used)

cybernetlab:
> However would using `:inet.setopts(sock, active: :once)` in the client process allow us to handle idle disconnect earlier and not have to retry a request? We should usually get a close notification message, and can reconnect proactively? Then instead of `:gen_tcp.recv` we could do `receive do {:tcp, ^socket, data} ->` etc.

Agree. The better solution is to handle disconnects at the moment of disconnect. I think that we can change active option to `false` right before `:gen_tcp.send` and switch it back to `:once` after `:gen_tcp.recv`. Is it reasonable approach?
 
> Also its not always the case that all connection closes can be retried, right now this could double the request rate to a badly behaving service :(.

Why it doubles request rate? If connection is closed at the moment of sending then where are no data sent at all and it will sent once after reconnect

lib/thrift/binary/framed/client.ex
cybernetlab: resolved in https://github.com/pinterest/elixir-thrift/pull/478/commits/6f97f5614fcaf28c03c094366f01b830a135168d

cybernetlab:
@fishcakez I'm ready to re-review )

lib/thrift/binary/framed/client.ex
fishcakez: The result is used as the return to `connect` but its calling the `handle_call` callback and matching on its result, which was able to return the connect tuple before. In that situation we would return `{:ok, s}` with wrong socket in the state. I'm not sure if its safe here or in future can to avoid matching on the full result so we can pass state through. We should always match on all results to be sure state isnt messed up.

lib/thrift/binary/framed/client.ex
```diff
@@ -286,15 +311,51 @@ defmodule Thrift.Binary.Framed.Client do
...
+          {:disconnect, error, s}
+        end
+
       {:error, _} = error ->
         {:disconnect, error, s}
     end
   end
 
+  @impl Connection
+  def handle_info({:tcp_closed, sock}, %{reconnect: true, sock: {_transport, sock}} = s) do
```
fishcakez: Handle `:ssl_closed` too? I think we can disconnect on tcp_error or ssl_error too. In fact if we get data we probably want to error to crash.

I think we want to disconnect on this always, as we are outside a request and so a disconnect allows us to reconnect.

lib/thrift/binary/framed/client.ex
```diff
@@ -286,15 +311,51 @@ defmodule Thrift.Binary.Framed.Client do
...
+
+  defp receive_message(:ssl, sock, timeout) do
+    receive do
+      {:ssl, ^sock, data} -> {:ok, data}
+      {:ssl_closed, ^sock} -> {:error, :closed}
+      {:ssl_error, ^sock, error} -> {:error, error}
+    after
+      timeout -> {:error, :timeout}
+    end
+  end
```
fishcakez: recv looks good 👌

fishcakez:
> Agree. The better solution is to handle disconnects at the moment of disconnect. I think that we can change active option to `false` right before `:gen_tcp.send` and switch it back to `:once` after `:gen_tcp.recv`. Is it reasonable approach?

I like what you did with `active: true`, left comment about always handling all socket messages. Outside a request any message is an issue and we could close.

> Why it doubles request rate? If connection is closed at the moment of sending then where are no data sent at all and it will sent once after reconnect.

Unfortunately with TCP we can not tell why the socket closed. There are three main scenarios:

1.  The socket closed before we sent the request, and we'd like to retry as we never go to send the request.

2. The socket closed at some point after we sent the request, and the server never received the full request so didn't try to handle it.

3. The socket closed after the server received the request. This is the case where we could increase the load on the server because the server could receive the request multiple times *but* the socket is closed before it can send a response. Most likely there is a bug in the server that causes the socket to get closed. A problematic example would be with the thrift server in this repo, and using Task.async inside the handler function. If the task processes crashes then the server processes will get an exit signal and exit - leading to the socket being closed without sending a response. As the client does not get a response it would resend the request, doubling the qps. This double qps occurence is quite concern to me.

I wonder if handling the disconnect immediately (active: true with message handling) is enough for the issue you're seeing as we would catch scenario 1) most of the time but never retry on 2) or 3). Perhaps we can split this PR in two, and follow up on reconnect retry in a second one if its still a problem.

cybernetlab:
> I wonder if handling the disconnect immediately (active: true with message handling) is enough for the issue you're seeing as we would catch scenario 1) most of the time but never retry on 2) or 3). Perhaps we can split this PR in two, and follow up on reconnect retry in a second one if its still a problem.

I'm writing this PR not only for my case but for any possible similar cases. But I agree that in case of 2) and 3) we should not resend data

cybernetlab:
@fishcakez I've kept `active: true` and remove resending logic. Ready to review

cybernetlab:
Hi, guys! I need a SASL authentication for hbase. I can prepare another PR with auth backend support, but I need this PR to be included in my code too. How can I achive this? Add my changes in this PR? Thanks!

lib/thrift/binary/framed/client.ex
```diff
@@ -291,10 +302,43 @@ defmodule Thrift.Binary.Framed.Client do
...
+  @impl Connection
+  def handle_info({:tcp_closed, sock}, %{reconnect: true, sock: {:gen_tcp, sock}} = s) do
+    {:disconnect, :reconnect, s}
+  end
+
+  def handle_info({:ssl_closed, sock}, %{reconnect: true, sock: {:ssl, sock}} = s) do
+    {:disconnect, :reconnect, s}
+  end
+
+  def handle_info(_, s) do
```
fishcakez: We should handle the `{:tcp_error, sock, _}`, `{:ssl_error, sock, _}`, `{:tcp, sock, _}, {;ssl, sock, _}` here but ignore messages from old sockets (or clean them up in disconnect/2).

cybernetlab:
@fishcakez ready to review )

cybernetlab:
@jparise is enough to add:

> The `:reconnect` option if setted to `true` forces client to reopen tcp connection whenever it closed.

as a last paragraph to the end of `@doc` for [`start_link/3`](https://github.com/pinterest/elixir-thrift/blob/master/lib/thrift/binary/framed/client.ex#L129)?

jparise:
> @jparise is it enough to add:
> 
> > The `:reconnect` option if setted to `true` forces client to reopen tcp connection whenever it closed.

`s/setted/set`

> as a last paragraph to the end of `@doc` for [`start_link/3`](https://github.com/pinterest/elixir-thrift/blob/master/lib/thrift/binary/framed/client.ex#L129) [](/pinterest/elixir-thrift/blob/HEAD@{2019-10-15T15:59:16Z}/lib/thrift/binary/framed/client.ex#L129)?

Yes, that would be fine.

lib/thrift/binary/framed/client.ex
```diff
@@ -124,6 +129,9 @@ defmodule Thrift.Binary.Framed.Client do
   Additionally, the options `:name`, `:debug`, and `:spawn_opt`, if specified,
   will be passed to the underlying `GenServer`. See `GenServer.start_link/3`
   for details on these options.
+
+  The `:reconnect` option if setted to `true` forces client to reopen tcp connection whenever
```
jparise:
```suggestion
  The `:reconnect` option if set to `true` forces client to reopen TCP connection whenever
```

cybernetlab: ) Already fixed - thanks!

lib/thrift/binary/framed/client.ex
```diff
@@ -291,10 +305,59 @@ defmodule Thrift.Binary.Framed.Client do
...
+  def handle_info({:ssl, sock, _data}, %{reconnect: true, sock: {:ssl, sock}} = s) do
+    {:disconnect, :reconnect, s}
+  end
+
+  def handle_info({:ssl_closed, sock}, %{reconnect: true, sock: {:ssl, sock}} = s) do
+    {:disconnect, :reconnect, s}
+  end
+
+  def handle_info({:ssl_error, sock, _error}, %{reconnect: true, sock: {:ssl, sock}} = s) do
+    {:disconnect, :reconnect, s}
```
fishcakez: I think we should always disconnect and make the decision on reconnect there. If `reconnect: false` we want to disconnect too.

cybernetlab: In this case many tests fails. If I disconnect on any TCP or SSL message with `reconnect: false` following output produced by `mix test`:

```
19:52 $ mix test
Compiling 1 file (.ex)
Excluding tags: [pending: true]

............................................................................................................................................

  1) test clients exit if they try to use a closed client (Thrift.Generator.ServiceTest)
     test/thrift/generator/service_test.exs:387
     ** (exit) exited in: :gen_server.call(#PID<0.1755.0>, {:call, "friend_ids_of", [<<10, 0, 1, 0, 0, 0, 0, 0, 0, 4, 210>> | <<0>>], []}, 5000)
         ** (EXIT) {:error, :closed}
     code: assert Client.friend_ids_of(client, 1234) == {:error, :closed}
     stacktrace:
       (stdlib) gen_server.erl:223: :gen_server.call/3
       (thrift) lib/thrift/binary/framed/client.ex:221: Thrift.Binary.Framed.Client.call/5
       test/thrift/generator/service_test.exs:392: (test)

     The following output was logged:
     
     19:52:36.114 [error] Connection closed
     
     19:52:36.115 [error] GenServer #PID<0.1755.0> terminating
     ** (stop) {:error, :closed}
     Last message: {:tcp_closed, #Port<0.19>}
     State: nil
     
..............

  2) test clients exit on connection timeout (Thrift.Generator.ServiceTest)
     test/thrift/generator/service_test.exs:373
     ** (exit) exited in: :gen_server.call(#PID<0.1958.0>, {:call, "friend_ids_of", [<<10, 0, 1, 0, 0, 0, 0, 0, 0, 4, 210>> | <<0>>], []}, 5000)
         ** (EXIT) no process: the process is not alive or there's no process currently associated with the given name, possibly because its application isn't started
     code: assert {:error, closed} = Client.friend_ids_of(client, 1234)
     stacktrace:
       (stdlib) gen_server.erl:223: :gen_server.call/3
       (thrift) lib/thrift/binary/framed/client.ex:221: Thrift.Binary.Framed.Client.call/5
       test/thrift/generator/service_test.exs:382: (test)

     The following output was logged:
     
     19:52:36.155 [error] Connection closed
     
     19:52:36.155 [error] GenServer #PID<0.1958.0> terminating
     ** (stop) {:error, :closed}
     Last message: {:tcp_closed, #Port<0.59>}
     State: nil
     
..

  3) test it handles void oneway functions (Thrift.Generator.ServiceTest)
     test/thrift/generator/service_test.exs:309
     ** (EXIT from #PID<0.1990.0>) {:error, :closed}

     The following output was logged:
     
     19:52:36.188 [error] Connection closed
     
     19:52:36.188 [error] GenServer #PID<0.1996.0> terminating
     ** (stop) {:error, :closed}
     Last message: {:tcp, #Port<0.68>, <<0>>}
     State: nil
     
     19:52:36.189 [error] GenServer #PID<0.1991.0> terminating
     ** (stop) {:error, :closed}
     Last message: {:EXIT, #PID<0.1990.0>, {:error, :closed}}
     State: {:state, {#PID<0.1991.0>, Supervisor.Default}, :one_for_one, {[ranch_listener_sup: Thrift.Generator.ServiceTest.ServerSpy], %{{:ranch_listener_sup, Thrift.Generator.ServiceTest.ServerSpy} => {:child, #PID<0.1992.0>, {:ranch_listener_sup, Thrift.Generator.ServiceTest.ServerSpy}, {:ranch_listener_sup, :start_link, [Thrift.Generator.ServiceTest.ServerSpy, :ranch_tcp, %{num_acceptors: 1, socket_opts: [port: 0]}, Thrift.Binary.Framed.ProtocolHandler, {Thrift.Generator.ServiceTest.SimpleService.Binary.Framed.Server, Thrift.Generator.ServiceTest.ServerSpy, [], []}]}, :permanent, :infinity, :supervisor, [:ranch_listener_sup]}}}, :undefined, 0, 5, [], 0, Supervisor.Default, {:ok, {%{intensity: 0, period: 5, strategy: :one_for_one}, [{{:ranch_listener_sup, Thrift.Generator.ServiceTest.ServerSpy}, {:ranch_listener_sup, :start_link, [Thrift.Generator.ServiceTest.ServerSpy, :ranch_tcp, %{num_acceptors: 1, socket_opts: [port: 0]}, Thrift.Binary.Framed.ProtocolHandler, {Thrift.Generator.ServiceTest.SimpleService.Binary.Framed.Server, Thrift.Generator.ServiceTest.ServerSpy, [], []}]}, :permanent, :infinity, :supervisor, [:ranch_listener_sup]}]}}}
     
.................

  4) test it can handle oneway messages (Servers.Binary.Framed.IntegrationTest)
     test/thrift/binary/framed/server_test.exs:158
     ** (EXIT from #PID<0.2192.0>) {:error, :closed}

     The following output was logged:
     
     19:52:36.324 [error] Connection closed
     
     19:52:36.324 [error] GenServer :"test it can handle oneway messages" terminating
     ** (stop) {:error, :closed}
     Last message: {:tcp, #Port<0.114>, <<0>>}
     State: nil
     
..................................................

Finished in 6.3 seconds
227 tests, 4 failures

Randomized with seed 350387
```

cybernetlab: I'm not sure - should I correct this test cases?

fishcakez:
```

Comment (cybernetlab) [reply lib/thrift/binary/framed/client.ex]: Hi, any comments about my last question?

Comment (fishcakez) [reply lib/thrift/binary/framed/client.ex]: We need to think about we handle these cases, and handle them.

For example we could do

```elixir
try do
  GenServer.call(...)
catch
  :exit, {{:error, _} = error, _} ->
    error
  :exit, {:noproc, _} ->
    {:error, :closed}
end
```

cybernetlab: My opinion is not to handle these cases and let users of library to handle it. But in this case some test should be rewritten

PR #498, opened 2020-03-25
Add Elixir 1.10 to the testing matrix

jparise:


jparise:
This change depends on #497.

PR #515, opened 2020-10-19
Eliminate redundant 'Elixir.' prefixes

pguillory:
Minor refactor for readability of compiled code.


jparise:
Does the example code need to be regenerated, too?

jparise (approved):

pguillory:
> Does the example code need to be regenerated, too?

Might have missed it earlier, got it now.

PR #516, opened 2020-10-19
Compile enums into case statements

pguillory:
Elixir is slow to compile functions with many heads, which is how we generate
enums in this library. It's faster to compile an equivalent case statement
inside a single function head. I observed that it reduced the time to compile
an enum with about 2500 values from 24s to 3s.

### DIFF
--- a/example/lib/calculator/generated/vector_product_type.ex
+++ b/example/lib/calculator/generated/vector_product_type.ex
@@ -8,28 +8,30 @@ defmodule(Calculator.Generated.VectorProductType) do
     2
   end
 
-  def(value_to_name(1)) do
-    {:ok, :dot_product}
-  end
-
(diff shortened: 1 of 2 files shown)


jparise:
Is there any runtime cost?

pguillory:
I benchmarked `Calculator.Generated.VectorProductType.value_to_name` before and after the change and found no performance difference.
```
## EnumBench
benchmark name          iterations   average time 
value_to_name - after     10000000   0.19 µs/op
value_to_name - before    10000000   0.19 µs/op
```

jparise (approved):

fishcakez:
During expansion from erlang to core erlang, the case and function heads go to equivalent (for what we care about here) AST. I remember we did this in Postgrex for similar optimization.

PR #517, opened 2020-10-20
Refactor Thrift.Transport.SSL.configuration/1

pguillory:
The implementation did not respect its type signature, which specified that the
error form of the return value should contain an exception as the second
element. This expectation is relied upon by maybe_ssl_handshake. However the
callback passed in the :configure option could return anything, and we return
its response without fully verifying that it matches the configuration
typespec.

Also refactored the code for clarity.

### DIFF
--- a/lib/thrift/transport/ssl.ex
+++ b/lib/thrift/transport/ssl.ex
@@ -33,7 +33,7 @@ defmodule Thrift.Transport.SSL do
           {:ok, opts} ->
             handle_optional(opts)
 
-          {:error, %_exception{}} = error ->
+          {:error, _} = error ->
             error
         end
 
--- a/lib/thrift/transport/ssl.ex
+++ b/lib/thrift/transport/ssl.ex
@@ -25,13 +25,19 @@ defmodule Thrift.Transport.SSL do
   @spec configuration([option]) ::
           {:required | :optional, [:ssl.ssl_option()]} | nil | {:error, Exception.t()}
   def configuration(opts) do
-    case Keyword.pop(opts, :enabled, false) do
-      {true, opts} ->
-        opts
-        |> handle_configure()
-        |> handle_optional()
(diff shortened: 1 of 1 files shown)


pguillory:
I must have closed #511 accidentally.

lib/thrift/transport/ssl.ex
```diff
@@ -25,13 +25,19 @@ defmodule Thrift.Transport.SSL do
...
+    {enabled, opts} = Keyword.pop(opts, :enabled, false)
 
-      {false, _} ->
+    case enabled do
+      true ->
+        case handle_configure(opts) do
+          {:ok, opts} ->
+            handle_optional(opts)
+
+          {:error, %_exception{}} = error ->
```
fishcakez, line 36: This is a little defensive as elixir patterns go because the second tuple would always match given the rest of the code. Usually wed only be so defensive at the edge of an API this defensively, which youve done below from the callback result.

lib/thrift/transport/ssl.ex
```diff
@@ -43,26 +49,23 @@ defmodule Thrift.Transport.SSL do
       {:ok, extra_opts} ->
         {:ok, extra_opts ++ opts}
 
-      {:error, _} = error ->
+      {:error, %_exception{}} = error ->
```
fishcakez, line 52: Perhaps `Exception.exception?` if we want to check this as an exception is more than `struct`.

fishcakez (approved):

lib/thrift/transport/ssl.ex
pguillory: Sure that's fine.

lib/thrift/transport/ssl.ex
pguillory: For now I've just matched the same pattern as in protocol_handler.ex and client.ex.

PR #519, opened 2020-11-04
Add server side metrics

pguillory:
We could use server-side metrics for Thrift services. Standard metrics for
every endpoint like QPS, latency, request/response size, and success rate.

In order to avoid coupling to our internal metrics library, I left the
interface pluggable. This also facilitates testing by substituting a stub
implementation in tests.

### DIFF
--- a/lib/thrift/binary/framed/protocol_handler.ex
+++ b/lib/thrift/binary/framed/protocol_handler.ex
@@ -280,24 +280,44 @@ defmodule Thrift.Binary.Framed.ProtocolHandler do
   defp start_span(metric_name, %__MODULE__{handler_module: handler_module}, tags \\ %{}) do
     event_name = [Thrift, handler_module, metric_name]
     time = System.monotonic_time()
-    metadata = Map.new(tags)
+    metadata = tags_to_metadata(tags)
     {:span, event_name, metadata, time}
   end
 
   defp finish_span({:span, event_name, metadata, time}, tags) do
     duration = System.monotonic_time() - time
-    :telemetry.execute(event_name, %{duration: duration}, add_tags(metadata, tags))
+    metadata = tags_to_metadata(tags, metadata)
+    :telemetry.execute(event_name, %{duration: duration}, metadata)
   end
 
   defp gauge(metric_name, value, %__MODULE__{handler_module: handler_module}, tags)
        when is_number(value) do
     event_name = [Thrift, handler_module, metric_name]
-    :telemetry.execute(event_name, %{value: value}, Map.new(tags))
+    metadata = tags_to_metadata(tags)
+    :telemetry.execute(event_name, %{value: value}, metadata)
   end
 
-  defp add_tags(metadata, tags) do
+  # Check that tags are simple and well-formed for use in systems like
+  # OpenTSDB. This check is only done in test due to the performance cost.
+  if Mix.env() == :test do
+    defp validate_tag(tag) when is_atom(tag) do
+      String.to_atom(validate_tag(Atom.to_string(tag)))
+    end
+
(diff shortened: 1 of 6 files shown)


jparise:
Should we consider using (or being compatible with) the [telemetry](https://hex.pm/packages/telemetry) package?

pguillory:
> Should we consider using (or being compatible with) the [telemetry](https://hex.pm/packages/telemetry) package?

Makes sense, I'll try that.

pguillory:
Updated to use telemetry.

lib/thrift/binary/framed/protocol_handler.ex
```diff
@@ -43,162 +51,257 @@ defmodule Thrift.Binary.Framed.ProtocolHandler do
...
+  end
+
+  defp close(%__MODULE__{transport: transport, socket: socket}) do
+    transport.close(socket)
+  end
+
+  defp start_span(metric_name, %__MODULE__{handler_module: handler_module}, tags \\ %{}) do
+    event_name = [Thrift, handler_module, metric_name]
+    time = System.monotonic_time()
+    metadata = Map.new(tags)
```
jparise, line 283: Do we need to make a copy of `tags`?

lib/thrift/binary/framed/protocol_handler.ex
```diff
@@ -43,162 +51,257 @@ defmodule Thrift.Binary.Framed.ProtocolHandler do
...
+    {:span, event_name, metadata, time}
+  end
+
+  defp finish_span({:span, event_name, metadata, time}, tags) do
+    duration = System.monotonic_time() - time
+    :telemetry.execute(event_name, %{duration: duration}, add_tags(metadata, tags))
```
jparise, line 289: Is `add_tags/2` here effective the same as `Map.merge(metadata, tags)`?

lib/thrift/binary/framed/protocol_handler.ex
pguillory: Tags is a keyword list when it is given as a parameter. Just for a little bit of readability really, to be able to write:
```
start_span(:call, state, method: method_name)
```
vs
```
start_span(:call, state, %{method: method_name})
```

lib/thrift/binary/framed/protocol_handler.ex
pguillory: Yeah tags is a keyword list, is the main thing.

pguillory (commented):
Telemetry metadata and "tags" are slightly different in subtle ways that are being glossed over here. Telemetry lets you put anything in metadata. However when we attach to these events in order to update metrics in OpenTSDB, the metadata is going to become tag on those metrics, which have particular requirements, like they need to be atoms/binaries with no whitespace or special characters. It would probably make sense to add a little validation to that effect.

lib/thrift/binary/framed/protocol_handler.ex
jparise: Ah, I missed that because of the default value for `tags` (which is `%{}`).

lib/thrift/binary/framed/protocol_handler.ex
jparise: We could also keep it a keyword list until this point, and then use `Map.new(Keyword.merge(metadata, tags))` here.

jparise (approved):

lib/thrift/binary/framed/protocol_handler.ex
pguillory: > Ah, I missed that because of the default value for `tags` (which is `%{}`).

Yeah that's confusing, will change.

PR #520, opened 2020-11-06
Add HTTP health check

pguillory:
Load balancers and service discovery mechanisms sometimes provide an HTTP
health check feature. We can auto-detect such an HTTP request being made to the
Thrift port and return a good enough response to pass the health check.

For convenience we can also print a little debugging information, so that if
you curl a port you can see what service is running there.

### DIFF
--- a/lib/thrift/binary/framed/protocol_handler.ex
+++ b/lib/thrift/binary/framed/protocol_handler.ex
@@ -4,6 +4,7 @@ defmodule Thrift.Binary.Framed.ProtocolHandler do
   @default_timeout 20_000
   @ssl_header_byte 0x16
   @tcp_header_byte 0x00
+  @http_header_byte 0x47
 
   @typedoc "A module that implements the :ranch_transport behaviour"
   @type transport :: :ranch_tcp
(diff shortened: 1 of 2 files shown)


jparise:
Is that a standard thing that Thrift servers provide? I don't think I've seen this before.

pguillory:
Yeah it's not standard, though neither is SSL auto-detection as far as I know.

An alternative would be to run a separate health check listener on a different port. It wouldn't be equivalent though, because the thrift listener might be having problems (like not accepting connections fast enough) while the health check listener stays up.

Another alternative would be to provide a TCP health check, for instance by echoing bytes, but it doesn't seem more standard, and a little less handy in terms of being able to curl a port.

A third alternative would be to just rely on the proxy checking if a TCP connection can be opened on the thrift port. No extra functionality to add here, but again it's not *quite* as good a health check, and a little less handy.

jparise:
My opinion is that this functionality should be provided by the application. It's in the best position to know the most useful kind of health response (content-wise) to return over which port(s)/protocol(s).

PR #521, opened 2020-11-06
Validate SSL configuration before starting server

pguillory:
If SSL certs are not available, then every SSL connection will fail. We should
check for this condition at startup before opening the port. That way if a new
node joins a cluster and its SSL configuration is not ready, it won't try to
start accepting connections that could have gone to a healthy host.

### DIFF
--- a/lib/thrift/binary/framed/server.ex
+++ b/lib/thrift/binary/framed/server.ex
@@ -1,6 +1,4 @@
 defmodule Thrift.Binary.Framed.Server do
-  require Logger
-
   @moduledoc """
   A server implementation of Thrift's Binary Framed protocol.
 
--- a/lib/thrift/binary/framed/server.ex
+++ b/lib/thrift/binary/framed/server.ex
@@ -1,4 +1,6 @@
 defmodule Thrift.Binary.Framed.Server do
+  require Logger
+
   @moduledoc """
   A server implementation of Thrift's Binary Framed protocol.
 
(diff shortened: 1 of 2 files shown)


lib/thrift/binary/framed/server.ex
```diff
@@ -78,4 +82,18 @@ defmodule Thrift.Binary.Framed.Server do
   def stop(pid) do
     Supervisor.stop(pid)
   end
+
+  # Ensure that SSL certs are available before starting server.
+  def validate_ssl_configuration!(ssl_opts) do
+    case Thrift.Transport.SSL.configuration(ssl_opts) do
+      {:error, %_exception{} = err} ->
+        Logger.error("Error validating SSL configuration: " <> Exception.format(:error, err, []))
```
jparise, line 90: Should this do more than log?

jparise (commented):
Consider adding a test case?

lib/thrift/binary/framed/server.ex
pguillory: Sure. Like raise `err`?

pguillory:
Added a unit test.

lib/thrift/binary/framed/server.ex
```diff
@@ -1,4 +1,6 @@
 defmodule Thrift.Binary.Framed.Server do
+  require Logger
```
jparise, line 2: I think this is now unused?

jparise (approved):

lib/thrift/binary/framed/server.ex
pguillory: Yep.

PR #534, opened 2021-01-20
Add connect_func option to server

pguillory:
We'd like to do application-level validation on a client's TLS cert. Currently
it lives within the connection process state, which is not accessible from a
handler function.

In order to make it accessible, this diff allows an optional connect_func
callback to be provided when starting a Thrift server. It will be called after
a TCP or SSL connection is established with the socket as its single parameter.
Given the socket, we can retrieve the client cert using :ssl.peercert/1, and
either do validation immediately or stash the cert in the process dictionary
for validation later in a handler function.

### DIFF
--- a/lib/thrift/binary/framed/protocol_handler.ex
+++ b/lib/thrift/binary/framed/protocol_handler.ex
@@ -24,6 +24,7 @@ defmodule Thrift.Binary.Framed.ProtocolHandler do
     :socket,
     :server_module,
     :handler_module,
+    :connect_func,
     :recv_timeout
   ]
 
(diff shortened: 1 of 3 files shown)


fishcakez:
PR looks good but maybe we can work out a clear/nicer name `connect_func`?

fishcakez (approved):

pguillory:
> PR looks good but maybe we can work out a clear/nicer name than `connect_func`?

Yeah it's not a good name. What would be better, `connect_cb`? `on_connect`?

fishcakez:
Sure `on_connect` or `after_connect` is similar to some other options in other libraries

pguillory:
Cool went with on_connect. Thanks @fishcakez.

jparise:
@pguillory it looks like this change didn't pass the CI build, and now the `master` branch's build is failing.

PR #564, opened 2021-11-14
Support THeader protocol

pguillory:
THeader is an optional wrapper for Thrift RPC messages that allows for requests
to include out-of-band headers, useful for things like distributed tracing in
service oriented architectures.

This change only implements THeader support on the server side. We can
auto-detect the protocol on a per-request basis based on the first two bytes.
THeader starts with 0x0FFF, whereas binary starts with 0x8001.

### DIFF
--- a/lib/thrift/protocol/theader.ex
+++ b/lib/thrift/protocol/theader.ex
@@ -1,0 +1,153 @@
+defmodule Thrift.Protocol.THeader do
+  alias Thrift.Protocol.Binary
+  alias Thrift.TApplicationException
+
+  # THeader format:
+  # https://github.com/apache/thrift/blob/master/doc/specs/HeaderFormat.md
+
+  @header_magic 0x0FFF
+
+  @binary_protocol 0
+  @compact_protocol 2
+
+  @zlib_transform 1
+  @hmac_transform 2
+  @snappy_transform 3
+
+  @info_keyvalue 1
+
+  # THeader format:
+  # https://github.com/apache/thrift/blob/master/doc/specs/HeaderFormat.md
+  def deserialize(
+        <<@header_magic::size(16), _flags::size(16), _sequence_number::size(32),
+          header_size::size(16), rest::binary>>
+      ) do
+    header_size_bytes = header_size * 4
+    <<header_binary::binary-size(header_size_bytes), payload::binary>> = rest
+    headers = deserialize_header(header_binary)
+
+    with {:ok, message} <- headers.protocol.deserialize(:message_begin, payload) do
+      {:ok, headers, message}
+    end
+  end
+
+  # Try other protocols.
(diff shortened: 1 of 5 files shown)


lib/thrift/protocol/theader.ex
```diff
@@ -0,0 +1,153 @@
...
+    end
+  end
+
+  # Try other protocols.
+  def deserialize(payload) do
+    with {:ok, message} <- Binary.deserialize(:message_begin, payload) do
+      headers = %{protocol: Binary}
+      {:ok, headers, message}
+    end
+  end
```
jparise, line 40: I'm not sure it feels quite right for the `THeader` module to be the thing that defers to the other protocols (like `Binary`). I think I would have expected the protocol handler to attempt to match on the magic header one level up and then invoke the appropriate protocol-specific deserialization calls from there.

lib/thrift/protocol/theader.ex
```diff
@@ -0,0 +1,153 @@
+defmodule Thrift.Protocol.THeader do
+  alias Thrift.Protocol.Binary
+  alias Thrift.TApplicationException
+
+  # THeader format:
+  # https://github.com/apache/thrift/blob/master/doc/specs/HeaderFormat.md
```
jparise, line 6: Maybe this should be part of a moduledoc block?

This link is also duplicated below (above `deserialize`).

jparise (commented):
This looks good overall, but I left a note about the general approach by which the protocol is detected and the protocol-specific code invoked.

It would also be good to add a short note to the `README.md` that mentioned `THeader` support.

lib/thrift/protocol/theader.ex
fishcakez: THeader is really a Transport and not a Protocol. Likely we want to move the abstraction to that level.

fishcakez:
If the server receives a THeader call, it needs to return a THeader reply, otherwise some clients may break. Generally speaking clients will have a configured a transport/protocol and always use it.

lib/thrift/protocol/theader.ex
pguillory: @jparise: With its protocol_id field, THeader knows about selecting between binary/compact protocols. So to me it feels like a natural place also to be doing protocol auto-detection.

@fishcakez: True that it's a transport. I'll rename it to Thrift.Transport.THeader.

lib/thrift/protocol/theader.ex
pguillory: Thanks, done.

pguillory:
> If the server receives a THeader call, it needs to return a THeader reply, otherwise some clients may break. Generally speaking clients will have a configured a transport/protocol and always use it.

Will work on this.

PR #590, opened 2022-12-15
Un-nest generated BinaryProtocol modules

pguillory:
A change in Elixir 1.13 prevents a nested module from referencing a struct defined by the parent module. We use this pattern in generated BinaryProtocol modules, which reference the struct defined by their parent, which corresponds to a Thrift struct.

Example code from the tests that won't compile in 1.13:

    defmodule Servers.Binary.Framed.IntegrationTest.IdAndName do
      @moduledoc false
      _ = "Auto-generated Thrift struct server_test.IdAndName"
      _ = "1: i64 id"
      _ = "2: string name"
      defstruct id: nil, name: nil
      @type t :: %__MODULE__{}
      def new do
        %__MODULE__{}
      end
      defmodule BinaryProtocol do
        @moduledoc false
        def deserialize(binary) do
          deserialize(binary, %Servers.Binary.Framed.IntegrationTest.IdAndName{})
        end

This diff eliminates the circular dependency by simply moving all of the serialize/deserialize implementation out of BinaryProtocol into the parent.

This will break client code that references the BinaryProtocol modules directly. However, the fix is a fairly straightforward search-and-replace.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,9 @@ jobs:
               elixir: 1.11
               otp: 23
               coverage: true
+          - pair:
+              elixir: 1.13
+              otp: 24
     env:
       MIX_ENV: test
     steps:
--- a/README.md
+++ b/README.md
@@ -154,9 +154,9 @@ your own types.
 ```elixir
 iex> alias Calculator.Generated.Vector
 iex> data = %Vector{x: 1, y: 2, z: 3}
-|> Vector.BinaryProtocol.serialize
+|> Vector.serialize
 |> IO.iodata_to_binary
-iex> Vector.BinaryProtocol.deserialize(data)
+iex> Vector.deserialize(data)
 {%Calculator.Generated.Vector{x: 1.0, y: 2.0, z: 3.0}, ""}
 ```
 
--- a/lib/thrift/binary/framed/client.ex
+++ b/lib/thrift/binary/framed/client.ex
@@ -333,9 +333,7 @@ defmodule Thrift.Binary.Framed.Client do
       TApplicationException.exception(
         type: :bad_sequence_id,
         message:
-          "Invalid sequence id. The client sent #{seq_id}, but the server replied with #{
-            mismatched_seq_id
-          }"
+          "Invalid sequence id. The client sent #{seq_id}, but the server replied with #{mismatched_seq_id}"
       )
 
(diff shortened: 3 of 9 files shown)


scohen:
Are you sure about this? I compiled and ran the following successfully under 1.13 (and 1.14)
```elixir
defmodule Servers.Binary.Framed.IntegrationTest.IdAndName do
  @moduledoc false
  _ = "Auto-generated Thrift struct server_test.IdAndName"
  _ = "1: i64 id"
  _ = "2: string name"
  defstruct id: nil, name: nil
  @type t :: %__MODULE__{}
  def new do
    %__MODULE__{}
  end

  defmodule BinaryProtocol do
    @moduledoc false
    def deserialize(binary) do
      deserialize(binary, %Servers.Binary.Framed.IntegrationTest.IdAndName{})
    end

    def deserialize(binary, struct) do
      IO.inspect(binary: binary, struct: struct)
    end
  end
end
```
```
Erlang/OTP 25 [erts-13.1.1] [source] [64-bit] [smp:12:12] [ds:12:12:10] [async-threads:1] [jit:ns]

Interactive Elixir (1.13.4) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> Servers.Binary.Framed.IntegrationTest.IdAndName.BinaryProtocol.deserialize("hello")
[
  binary: "hello",
  struct: %Servers.Binary.Framed.IntegrationTest.IdAndName{id: nil, name: nil}
]
```

I'm pursuing a similar strategy in `elixir_ls` and am not having any problems with nested modules referring to a parent struct and haven't encountered any trouble in any version of elixir since 1.10. The above fails if you just pass `IdAndName` as the name of the parent struct without aliasing. 
Maybe I’m not seeing the circular dependency. The example above didn’t include one.

pguillory:
Thanks for catching Steve. It seems like the problem I encountered was with the combination of Elixir 1.13.0 and Erlang 24.1.7 that was installed by Homebrew on my laptop. Ubuntu refused to install that combination citing a requirement for Erlang >=25, so maybe they're incompatible? Anyway upgrading fixed the issue, so I'll close this.

scohen:
That's really interesting @pguillory. I hadn't tried that combination. 
btw, i've moved to asdf for managing elixir and many other languages as opposed to homebrew. It's a much better solution, especially if you want to have multiple installs and switch between them.

PR #591, opened 2022-12-15
Support new Ranch options syntax

pguillory:
Starting with Ranch 2.0, you get a warning when mixing Ranch and TCP socket options in a single keyword list. We only support the old syntax, so you'll get warnings at startup, and presumably that syntax will stop working at some point.

This diff adds support for the new syntax. It does so in a backwards compatible way, so you can upgrade Ranch and your option syntax either before or after upgrading elixir-thrift.

### DIFF
--- a/lib/thrift/binary/framed/server.ex
+++ b/lib/thrift/binary/framed/server.ex
@@ -54,7 +54,7 @@ defmodule Thrift.Binary.Framed.Server do
     transport_opts =
       opts
       |> Keyword.get(:transport_opts, [])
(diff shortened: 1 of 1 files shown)


jparise (approved):
Looks good. Would also be nice to have some unit test coverage for both paths to avoid regressions.

pguillory:
I'll go ahead and merge this but let's talk test coverage. Seems like the ideal, as a library that depends on another library (Ranch), the ideal would be to test against multiple versions of Ranch. Is there a common pattern for doing that?

PR #592, opened 2022-12-16
Fix CI tests failing on ubuntu-latest

pguillory:
Tests on other PRs are failing with the following error:

    Error: Requested Erlang/OTP version (23) not found in version list (should you be using option 'version-type': 'strict'?)

I tried changing the Ubuntu version used in the CI environment based on Googling for ideas, and it seems to get the tests running.


jparise (approved):

pguillory:
I'm guessing they've been failing since the Ubuntu 22.10 release in October. Maybe it doesn't support older versions of Erlang/OTP?

